PR #62, opened 2015-01-06
Adds UnitNode support (same as Scalatags) back.

matthughes:
Allows partial matches when building your VDOM.  This worked before the Scalatags embedding and presumably just missed along the way:

```
div(
  (if (someThingEnabled) span("some thing")),
  div("always here")
)
```

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/vdom/Scalatags.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/vdom/Scalatags.scala
@@ -266,6 +266,14 @@ private[vdom] object Scalatags {
   implicit class SeqNode[A <% TagMod](xs: Seq[A]) extends TagMod {
     def applyTo(t: Builder) = xs.foreach(_.applyTo(t))
   }
+ 
+  /** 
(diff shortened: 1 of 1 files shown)


japgolly:
Sorry you took the time to submit a patch but no, `Unit` more often than not an in indication of side-effects. Having an implicit conversion to silently gobble it up is dangerous, especially with non-Scalaz callbacks (which resolve to `Unit` when invoked) around. There are safer ways for your example (below).

From changelog:
* Removed Scalatags' `()` to `Tag` implicit conversion. It seems like a good idea but can cause certain issues (esp with type inference) to be silently ignored, and is unsafe. Use `EmptyTag` instead.
* An unintended consequence, `bool && (attr := value)` changed to `bool ?= (attr := value)`.
* Attributes, styles, tags, modifiers support being wrapped in `Option`, `js.UndefOr`, or `scalaz.Maybe`.

japgolly:
To be super clear:
```scala
if (someThingEnabled) span("some thing")
```

can be:
```scala
someThingEnabled ?= span("some thing")
```

matthughes:
Ah great, thanks.  Sorry I missed that in the changelog.

PR #65, opened 2015-01-12
ReactCloneWithProps addon and isMounted guard for async calls

chandu0101:
http://facebook.github.io/react/docs/clone-with-props.html

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/React.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/React.scala
@@ -176,7 +176,7 @@ trait ComponentScope_S[+State] extends Object {
 trait ComponentScope_SS[State] extends ComponentScope_S[State] {
   @JSName("setState") def _setState(s: WrapObj[State]): Unit = ???
   @JSName("setState") def _setState(s: WrapObj[State], callback: UndefOr[JFn]): Unit = ???
-  @JSName("isMounted") def _isMounted() : Boolean = ???
+  @JSName("isMounted") def isMounted() : Boolean = ???
 }
 
 trait ComponentScope_B[+Backend] extends Object {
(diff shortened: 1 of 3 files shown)


core/src/main/scala/japgolly/scalajs/react/React.scala
```diff
@@ -176,6 +176,7 @@ trait ComponentScope_S[+State] extends Object {
 trait ComponentScope_SS[State] extends ComponentScope_S[State] {
   @JSName("setState") def _setState(s: WrapObj[State]): Unit = ???
   @JSName("setState") def _setState(s: WrapObj[State], callback: UndefOr[JFn]): Unit = ???
+  @JSName("isMounted") def _isMounted() : Boolean = ???
```
japgolly: We don't need the underscore here because we don't need to transform it before we make it available to Scala. Just `def isMounted: Boolean` is fine. (Also no `()` because it's not side-effecting.)

japgolly:
@chandu0101 Could you split this into 2 PRs? One for `isMounted` and the other for `ReactCloneWithProps`? Also could you also just chuck a quick test in for `ReactCloneWithProps`?

core/src/main/scala/japgolly/scalajs/react/React.scala
chandu0101: if i remove ()  , i am getting following error
```scala
 caused by: java.lang.ClassCastException: function () {
[info]         return ReactComponent.Mixin.isMounted.call(this) &&
[info]           this._compositeLifeCycleState !== CompositeLifeCycle.MOUNTING;
[info]       } is not an instance of java.lang.Boolean
```

japgolly: Oh sorry, ignore me then :grinning:

chandu0101: :) , btw how can i split into two PR's ? ,Could you please link me to some doc ..
how about gitter for scalajs-react ?( useful for people like me to ask stupid questions :P )

japgolly: You use different git branches. But I'm hungry so don't worry, 1 PR is fine today. And yeah re: gitter but later. For now: _foooooooooooooooooood_!

chandu0101:
added  a simple test to ReactCloneWithProps

japgolly:
thanks mate

PR #70, opened 2015-01-19
Change attribute names to reactjs naming and add spreadMethod

lucidd:
This fixes react ignoring those attributes and printing warnings like `Warning: Unknown DOM property stroke-opacity. Did you mean strokeOpacity?`.
The attribute names are adjusted according to [this](https://github.com/facebook/react/blob/master/src/browser/ui/dom/SVGDOMPropertyConfig.js#L67-L86).


japgolly:
Thanks and sorry for the delay.

PR #73, opened 2015-02-10
Some minor changes and new styles

lucidd:


core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
```diff
@@ -2467,7 +2467,46 @@ trait HtmlStyles {
...
+   * Sometimes it is of interest to let the outline of an object keep its original width no matter which transforms are
+   * applied to it. For example, in a map with a 2px wide line representing roads it is of interest to keep the roads
+   * 2px wide even when the user zooms into the map. To achieve this, SVG Tiny 1.2 introduces the 'vector-effect'
+   * property. Future versions of the SVG language will allow for more powerful vector effects through this property
+   * but this version restricts it to being able to specify the non-scaling stroke behavior
+   *
+   * w3.org
+   *
+   */
+  object vectorEffect extends Style("vectorEffect", "vectorEffect") {
```
japgolly: This should be in SvgAttrs instead of HtmlAttrs.

core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
```diff
@@ -2981,6 +3020,217 @@ trait HtmlStyles {
...
+   * MDN
+   */
+  object alignItems extends Style("alignItems", "alignItems") {
+
+    /**
+     * The cross-start margin edge of the flex item is flushed with the cross-start edge of the line.
+     *
+     * MDN
+     */
+    final val flexStart = this := "flex-start"
```
japgolly: Does React not expect CSS *values* in camel-case as well? Could you use this both as is and in camel-case and see if React a) issues a warning, b) generates the right CSS?

core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
```diff
@@ -2981,6 +3020,217 @@ trait HtmlStyles {
...
+     * MDN
+     */
+    final val row = this := "row"
+
+    /**
+     * Behaves the same as row but the main-start and main-end points are permuted.
+     *
+     * MDN
+     */
+    final val rowReverse = this := "row-reverse"
```
japgolly: `final val` is final for the CSS key declarations but could you please change all the K/Vs
```scala
final val xxx = this := "xxx"
```
to
```scala
@inline final def xxx = this := "xxx"
```

japgolly:
Finally, could you please update the changelog in `doc/`.
Quick and easy is fine, like:
```md
* Added new attributes:
  * `flex`
  * `george`
  * `etc`
```

core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
lucidd: I tested it and it works the way it is now (is there a name for that case?). When using a camel-case value React just passes the wrong value through without complaining.

core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
lucidd: ok i will do that but why def instead of val? (just wondering)

lucidd:
i just realized that removing those unused type parameter would probably be considered a breaking change wouldn't it? I will redo this PR without that change and submit a separate PR for later merging.

core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
lucidd: Generally vector-effect can be used as attribute or as a style but for some reason it does not work when using it as attribute with react (probably because its not on the [supported attribute list](http://facebook.github.io/react/docs/tags-and-attributes.html)). When using it as a style it works fine. So should i put it in SvgAttr, HtmlStyles or create a new file for SvgStyles?

core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
japgolly: Awesome, thanks for confirming that.

core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
japgolly: `val` stores the value in memory (when the object is initialised). For efficiency, all the tags and attribute keys are `val`s but the `this := value` shortcuts I decided to make inline aliases, rather than caching. It matches the way you normally use attributes.

japgolly:
It shouldn't be a breaking change to remove those unused type params because IIRC those classes are private and can be considered implementation details. I was gonna let it go but I would prefer that change in a separate patch so I'm good that it's removed :)

core/src/main/scala/japgolly/scalajs/react/vdom/SvgStyles.scala
```diff
@@ -0,0 +1,47 @@
...
+   * Sometimes it is of interest to let the outline of an object keep its original width no matter which transforms are
+   * applied to it. For example, in a map with a 2px wide line representing roads it is of interest to keep the roads
+   * 2px wide even when the user zooms into the map. To achieve this, SVG Tiny 1.2 introduces the 'vector-effect'
+   * property. Future versions of the SVG language will allow for more powerful vector effects through this property
+   * but this version restricts it to being able to specify the non-scaling stroke behavior
+   *
+   * w3.org
+   *
+   */
+  object vectorEffect extends Style("vectorEffect", "vectorEffect") {
```
japgolly: Looking at the http://www.w3.org/TR/SVGTiny12/painting.html and http://www.w3.org/TR/SVGTiny12/examples/non-scaling-stroke.svg, this just be a normal attribute. SVG doesn't have styles.

So `final val vectorEffect = "vector-effect".attr` or probably `final val vectorEffect = "vectorEffect".attr` if React complains and in the `SvgAttrs` file.

japgolly: Just saw your explanation in email (although I can't see it here on GH), so React ignores it as an attribute but it works as a style? Geez... Ok in that case could you please just put it in SvgAttr but with a note explaining why it's a `Style`.

lucidd: Github hides comments on outdated diffs by default. Most SVG attributes can also be used from CSS so maybe in the future there is a place for SvgStyles but i put in SvgAttr for now :). I will do the `@inline final def` in a separate PR so it doesn't get any messier in this one.

lucidd:
Ok i think thats it now. I published it locally and tested it with my project which uses most of the styles i added.

japgolly:
Merged, thanks!

PR #74, opened 2015-02-11
Remove unused type params

lucidd:
AutoStyle etc don't seem to be private so i think its still a braking change. I'll just leave that here so we don't forget it in the future. :)


japgolly:
Cool thanks.

japgolly:
Merged in 6a4c3dd2f1af20e2abe08e010f28cef2de4a31f0.

PR #75, opened 2015-02-11
Change from final val to @inline final def for this := ...

lucidd:
I hope i got all of them if not just let me know.


japgolly:
Wow, I didn't realise there were so many! I thought I did this before. I hope you regex'd it. Thanks a lot mate. :heart:

PR #93, opened 2015-03-06
Change SetInterval to take FiniteDuration instead of Double.

matthughes:
I vote for changing method signature, but we could also add this in *addition* to the Double method.


japgolly:
Thanks @matthughes !

PR #97, opened 2015-03-21
Change jquery call to use jQuery, not the Dynamic.global

ericssmith:
Code doesn't run as is. Changed to use jQuery.

### DIFF
--- a/gh-pages/src/main/scala/ghpages/examples/PictureAppExample.scala
+++ b/gh-pages/src/main/scala/ghpages/examples/PictureAppExample.scala
@@ -2,7 +2,9 @@ package ghpages.examples
 
 import japgolly.scalajs.react._, vdom.all._
 import scala.scalajs.js
+import org.scalajs.jquery._
 import ghpages.examples.util.SideBySide
+
 
 /**
  * Created by chandrasekharkode on 11/18/14.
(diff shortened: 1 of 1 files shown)


japgolly:
@chandu0101 ?

chandu0101:
hmm ..
```js
<script>
    function jsonp(url, callback) {
        var callbackName = 'jsonp_callback_' + Math.round(100000 * Math.random());
        window[callbackName] = function(data) {
            delete window[callbackName];
//            document.body.removeChild(script);
            callback(data);
        };
        var script = document.createElement('script');
        script.src = url + (url.indexOf('?') >= 0 ? '&' : '?') + 'callback=' + callbackName;
        document.body.appendChild(script);
    }
</script>
```

in gh-pages we have this in index.html , only for jsonp i didn't feel like adding library dependency at that time! ,feel free to change :)

japgolly:
Ah ok, I understand this a bit more now.

Hmmm, ok I think what we should do is move this jsonp method out of JS in index.html into a Scala.JS function and then put it in the PictureExample with a note that JQuery provides it too.

Or something similar to that. Pretty much exactly what you said @chandu0101 except not adding it as part of scalajs-react itself, just part of the example. And if that's what you already meant (I just reread and it could go either way) then I agree :D Clearly I'm a little tired right noww. :P

japgolly:
Thanks for this PR but #320 will supersede this.

PR #99, opened 2015-03-28
Added Touch event attributes and example with Touch events

aurelijusb:
Touch events as described in:
https://facebook.github.io/react/docs/events.html

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -25,6 +25,7 @@ object ScalajsReact extends Build {
         scalacOptions     ++= Seq("-deprecation", "-unchecked", "-feature",
                                 "-language:postfixOps", "-language:implicitConversions",
                                 "-language:higherKinds", "-language:existentials"),
+        scalacOptions     ++= Seq("-Xmax-classfile-name", "100"),
         updateOptions      := updateOptions.value.withCachedResolution(true))
 
   def preventPublication: PE =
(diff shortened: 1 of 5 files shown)


project/Build.scala
```diff
@@ -25,6 +25,7 @@ object ScalajsReact extends Build {
         scalacOptions     ++= Seq("-deprecation", "-unchecked", "-feature",
                                 "-language:postfixOps", "-language:implicitConversions",
                                 "-language:higherKinds", "-language:existentials"),
+        scalacOptions     ++= Seq("-Xmax-classfile-name", "100"),
```
japgolly: What's this for?

japgolly:
Cooool. :+1: Could you please move the changelog bit down into 0.8.3's section. Only 0.8.2 has been released so far so 0.8.3 is the bucket this patch will go into.

project/Build.scala
aurelijusb: It is written in comit message https://github.com/aurelijusb/scalajs-react/commit/ecbd83b3d4f44bf6f83eb58d670dad7ef0b6917a

On my Ubuntu machine I have encrypted home folder.
During compilation Scala creates very long class names and I get
"File name too long" exception.
So this is a small hack to make compilation work on encrypted file system.

aurelijusb:
`changelog` updated.

project/Build.scala
japgolly: Ah thanks.

Ok so have a read of this: http://scala-language.1934581.n4.nabble.com/Xmax-classfile-name-td4637574.html

This option seems quite dangerous. It affects binary compatibility and needs all dependencies to have the same setting, else :boom:.

Let's remove this and get the rest of the patch in.

aurelijusb:
Reverted `Xmax-classfile-name` commit. Now only TouchEvent chagnes are left.

japgolly:
Awesome, thanks for the patch.

PR #105, opened 2015-04-21
Allow scalatags to accept js.Array as an attribute value

codingismy11to7:
As stated at http://facebook.github.io/react/docs/forms.html, `value` / `defaultValue` of `multiple` `select` elements can be arrays, so I added support. Working properly on a local build.

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/vdom/Implicits.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/vdom/Implicits.scala
@@ -35,6 +35,9 @@ abstract class Implicits extends LowPri {
   @inline implicit final def _react_attrOptional[T[_], A](implicit t: Optional[T], a: AttrValue[A]): AttrValue[T[A]] =
     new OptionalAttrValue[T, A](t, a)
 
+  @inline implicit final def _react_attrArray[A <% js.Any]: AttrValue[js.Array[A]] =
+    new ArrayAttr[A]
+
   // Styles
   @inline implicit final def _react_styleString : StyleValue[String]  = stringStyleX
           implicit final val _react_styleBoolean: StyleValue[Boolean] = GenericStyle.stringValue
--- a/core/src/main/scala/japgolly/scalajs/react/vdom/Scalatags.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/vdom/Scalatags.scala
@@ -223,6 +223,10 @@ private[vdom] object Scalatags {
     @inline def apply[T <% js.Any] = new GenericAttr[T](a => a)
   }
 
+  final class ArrayAttr[T] extends AttrValue[js.Array[T]] {
+    def apply(v: js.Array[T], b: js.Any => Unit): Unit = b(v)
+  }
+
   final class GenericStyle[T](f: T => String) extends StyleValue[T] {
     def apply(b: Builder, s: Style, v: T): Unit = {
       b.addStyle(s.cssName, f(v))
--- a/test/src/test/scala/japgolly/scalajs/react/CoreTest.scala
+++ b/test/src/test/scala/japgolly/scalajs/react/CoreTest.scala
@@ -3,7 +3,7 @@ package japgolly.scalajs.react
 import japgolly.scalajs.react.Addons.ReactCloneWithProps
 import utest._
 import scala.scalajs.js, js.{Array => JArray}
-import org.scalajs.dom.raw.HTMLInputElement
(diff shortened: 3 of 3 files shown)


core/src/main/scala/japgolly/scalajs/react/vdom/Scalatags.scala
```diff
@@ -223,6 +223,13 @@ private[vdom] object Scalatags {
     @inline def apply[T <% js.Any] = new GenericAttr[T](a => a)
   }
 
+  final class ArrayAttr[T] extends AttrValue[js.Array[T]] {
+    def apply(v: js.Array[T], b: js.Any => Unit): Unit = b(v)
+  }
+  object ArrayAttr {
```
japgolly: No need for this object; it's not being used, nor is it visible.

core/src/main/scala/japgolly/scalajs/react/vdom/Scalatags.scala
```diff
@@ -223,6 +223,13 @@ private[vdom] object Scalatags {
     @inline def apply[T <% js.Any] = new GenericAttr[T](a => a)
   }
 
+  final class ArrayAttr[T] extends AttrValue[js.Array[T]] {
```
japgolly: `T` should be constrained as `T <: js.Any`

japgolly: Actually arrays are covariant so no need for a type param. Just make it accept `js.Array[js.Any]`

core/src/main/scala/japgolly/scalajs/react/vdom/Implicits.scala
```diff
@@ -35,6 +35,9 @@ abstract class Implicits extends LowPri {
   @inline implicit final def _react_attrOptional[T[_], A](implicit t: Optional[T], a: AttrValue[A]): AttrValue[T[A]] =
     new OptionalAttrValue[T, A](t, a)
 
+  @inline implicit final def _react_attrArray[A <% js.Any]: AttrValue[js.Array[A]] =
```
japgolly: I'm not sure how I feel about using a view bound here... If the view bound is used, `Array[A]` will need to be converted to `Array[_ <: js.Any]` before being passed to React (as you'll see anyway when you change the `ArrayAttr` type). If `A` already extends `js.Any` then you create a new array for no reason (or mutate the input = :grimacing:)

core/src/main/scala/japgolly/scalajs/react/vdom/Scalatags.scala
codingismy11to7: I'll give that a try

core/src/main/scala/japgolly/scalajs/react/vdom/Scalatags.scala
codingismy11to7: I just added it to match your style, and I was going to use it in the Implicit object, but didn't (to match your style)

core/src/main/scala/japgolly/scalajs/react/vdom/Scalatags.scala
codingismy11to7: Nope, I couldn't get it to work any other way. Not smart enough. I barely understand covariance and contravariance. In this case, React wants an array of strings, scala.js has an implicit conversion from string to js.Any, but String is not an Any. Unless I need to take an extra step of making sure my array is an array of Any...

Sorry, I just don't understand

japgolly:
Hey I'll have to reply to you later, friend. There isn't a test case with this. How about you leave the implementation for now and write up a quick test case to ensure & demonstrate it works as expected?

(`test` is SBT will give an error about the `gh-pages` module. Ignore it. You can use `test/test` or the alias `t` instead. I normally use `~t`.)

codingismy11to7:
Ok, I'll try to get to that this week - I'll take out the object and add a test and amend the commit...once I amend I'll come back and leave a comment, or feel free to close the PR and I'll open a new one (not positive when I'm going to have time to do this, Monday or Tuesday next week at the latest).

Thanks!

codingismy11to7:
Ok, I removed the unused ArrayAttr object, and added a test case. Let me know what else I can fix up - no hurry, we're using a copy in our codebase.

japgolly:
Thanks. It turns out it was talking complete rubbish about arrays being covariant; they're invariant. Sorry for causing confusion.

codingismy11to7:
No problem, thanks a lot

PR #106, opened 2015-04-22
ref to third party components

chandu0101:


japgolly:
Can you please put together a quick test that shows how this works (and of course confirms that it does work).

chandu0101:
added test .

japgolly:
Merged via 8b08610b0508fcbeb0c1227258e50a3a6fabf913.

Notes:
* If you create patches on different branches (on your side), you don't need to wait for me to accept a PR before you send me the next one ;)
* Is there really a need for `ReactCssTransitionGroupM`? I don't think there is and removing it, I was able to remove all of your changed to Refs as well. Have a look at e18727cbaefd1d4a2f5ae173164cd1b54d0b34c0 and let me know if there's any problem.

chandu0101:
hmm , ReactCssTransitionGroupM   should be there    , which is just facade for list of mounted methods of component  . Current Ref simple is bounded till TopNode , but to define facade we extends js.Object not TopNode !

** *And I don't understand why you deleted/merged/closed without discussion  mate :s ***

japgolly:
I closed the pull request because the pull request has been merged. (I had to merge and close via the command-line rather that click  'merge' in GitHub to shift the target branch.)

I'm not in the habit of immediately blocking people after I merge their requests so we'll notice that we're actually having discussion. Many would even would interpret the 7 lines of text I wrote to you above as discussion; I do.

Seeing as you're online let's continue the discussion on gitter. I'll open it now.

PR #123, opened 2015-05-19
change TopNode from dom.html.Element to dom.Element

james-za:
React has decent support for svg components. Setting TopNode to dom.Element would allow properly typed getDOMNode access of svg elements (in both component building and refs), given that its only subtypes are HTMLElement and SVGElement.

As an aside: className seems to have been misplaced in the dom facade, requiring a small workaround (it should be in dom.Element according to https://developer.mozilla.org/en-US/docs/Web/API/Element/className).

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/package.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/package.scala
@@ -6,7 +6,7 @@ import js.{Dynamic, UndefOr, undefined, Object, Any => JAny, Function => JFn}
 
 package object react {
 
-  type TopNode = html.Element
+  type TopNode = dom.Element
 
   type ReactEvent            = SyntheticEvent           [dom.Node]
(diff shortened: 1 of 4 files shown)


japgolly:
Cool, thanks. This would be fine if not for the dom facade `className` issue which bothers me slightly... Will you be raising a PR in scala-js-dom to fix it? If you do, maybe we can get Seb to cut a new release and avoid the workaround here. (FYI I plan to release 0.9.0 of this by the end of the month.)

james-za:
After checking the official DOM spec, this seems to be incorrect information from MDN.

I've refactored out the "matching a predicate as subtype" to clean up what had used js.Dynamic, the alternative being `n.isInstanceOf[HTMLElement] && n.asInstanceOf[HTMLElement].className`.

japgolly:
Alright, that'll work. Would you mind closing this and resubmitting it against the 0.9 branch pls? Github won't let me change the target myself.

PR #126, opened 2015-05-22
update refs example with new scalajs-dom type

chandu0101:


japgolly:
I appreciate the PR but if a new user uses `HTMLInputElement` not knowing where it comes from, IDEs can figure it out and auto-import. If we change it to `html.Input` the user will need to know specifically which import is needed and add the import manually. Better to leave this as is.

chandu0101:
cool okay :)

PR #128, opened 2015-05-26
Load react lib over proper protocol

zlangbert:
Makes sure when loading GitHub pages over https the browser doesn't block mixed content

### DIFF
--- a/gh-pages/index.html
+++ b/gh-pages/index.html
@@ -6,7 +6,7 @@
     <link rel="stylesheet" href="res/bootstrap.css">
     <link rel="stylesheet" href="res/app.css">
     <link rel="stylesheet" href="res/default.css">
-    <script type="text/javascript" src="http://cdnjs.cloudflare.com/ajax/libs/react/0.12.1/react-with-addons.min.js"></script>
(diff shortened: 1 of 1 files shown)


zlangbert:
I can no longer view any of http://japgolly.github.io/scalajs-react/. I've been able to before so I'm not sure what changed. Maybe default browser settings have changed or it could be an addon. Either way, loading resources over the same protocol as the page is usually what you want to do.

japgolly:
Thanks, merged in c91df56d2fcd49f00af5943b91164f22c98baaf1.

gh-pages are updated, can you see https://japgolly.github.io/scalajs-react/ now?

zlangbert:
Yup! All good now, thanks

japgolly:
Awesome, thank *you*. I'll also go around and update links to it to use https.

elacin:
I don't know if that matters to anybody, but this broke running the examples without a server - through file:///...

japgolly:
@elacin I normally run `gh-pages/serve.sh` + `sbt '~;clear;gh-pages/fastOptJS`  and view it through `localhost:4000`

PR #129, opened 2015-05-27
 Upgrade Scala.JS to 0.6.3

chandu0101:


japgolly:
What's the advantage here?

If we stay on 0.6.2 it means that scalajs-react will work with any 0.6.x version of Scala.JS including 0.6.3.

chandu0101:
indeed , but why not update ?

japgolly:
If we update it means that scalajs-react will never run on Scala.JS 0.6.{0,1,2}. That's not a big deal so I wouldn't care if there were something to be gained in exchange, **BUT**, here there seems to be no benefit.

In other words, I'd pay some freedom for something but I don't see why we should pay for nothing.

chandu0101:
i am confused here , how come 0.6.2  works fine on 0.6.0/1 but 0.6.3 not on .2/.1/.0 :S

japgolly:
The version of their IR (like SJS bytecode) is what determines the backwards compatibility and they don't always increase it with each SJS release. In SJS versions 0.6.0 - 0.6.2 it stayed the same but then they bumped it up in 0.6.3 for some fix. So while we stick to compiling it with 0.6.2 it will work with any 0.6.x release of SJS.

chandu0101:
ic thanks for info :) , anyway i prefer to update version since nothing stops old users(0,1,2) to upgrade to 0.6.3! .

japgolly:
np :) Regarding the change though, like I said, why pay something for nothing. As far as I understand, this change does exactly one thing: restrict potential users. No benefit.

chandu0101:
hmm , whats the best time to update then ? . Its been 16 days since 0.6.3 released and 0.9 is due in 3 days ..

japgolly:
It's not about when; it's about why. Give me a reason to update (specifically: a penalty for not updating) and I will.

japgolly:
(Note: clients are free to update to 0.6.3 - I use 0.6.3 on my work project with this.)

chandu0101:
cool okay ;)

japgolly:
If you discover a penalty for not updating, let me know - we'll reopen this.

japgolly:
..and only two hours later a need presents itself: Travis CI needs it.

We're now on 0.6.3 via 837ed50241b92e7a11f6f78b3d379afe84bcdf2b

Issue #134, opened 2015-05-31
Unable to create event listeners for dblclick


elacin:
I've pretty much tried everything, but I'm not able to get any element to respond to a double click.

I managed to create a failing test against the simulator, check the following commit:
https://github.com/elacin/scalajs-react/commit/ddef1ef558d5eb5afec95f20c8882447285d2aba

It succeeds for every other event type. I will PR it if you would want that test for later.

japgolly:
It turns out React use `onDoubleClick` instead of `onDblClick`.

Workaround:
```scala
val onDblClick = "onDoubleClick".reactAttr
```

> It succeeds for every other event type. I will PR it if you would want that test for later.

Good to know the rest are ok, and yes please, that test PR would be greatly appreciated.

elacin:
Ah wonderful - that did it! :)

Sending it now, then

PR #135, opened 2015-05-31
test event handling

elacin:
ref #134


japgolly:
Brilliant, thanks @elacin!

Issue #116, opened 2015-05-04
Issues with overloaded methods (render)


elacin:
I'm having a lot of fun hacking on scalajs-react, and wanted to raise one issue that keeps bothering me.
In most examples the following style is used:

```scala
    ReactComponentB[List[String]]("app")
      .initialState(State(0))
      .backend(new Backend(_))
      .render((P, S, B) ⇒ {
        val a = S.index
        div(s"State: $a")
      }
      ).build

```

Personally, i always have the urge to rewrite that into the following to get better pattern matching capabilities, and to avoid double set of parentheses if there is more than one statement within the function
```scala
  ReactComponentB[List[String]]("app")
    .initialState(State(0))
    .backend(new Backend(_))
    .render {
      case (_, state, _) ⇒
        val a = state.index
        div(s"State: $a")
    }.build

```

This, however, doesn't work, because of the multiple overloads of `render()`, and scalac spits out 
```
Error:(56, 13) missing parameter type for expanded function
The argument types of an anonymous function must be fully known. (SLS 8.5)
Expected type was: ?
    .render {
            ^
```

So this is, as far as i can see, an unfortunate limitation of the compiler that makes the library a bit confusing to use. Would you consider a PR where I either change the method names or (I'm sure preferably) introduce new wrapper functions so it's easy to disambiguate?

If yes, any naming preference?
- `render`, `renderWithChildren`, `renderComponentScope` 
- or shorter versions like for example `render`, `renderC`, `renderCS`

japgolly:
Hi. Awesome to hear you're having fun. Adding `case` doesn't  make much sense though. You use it to match patterns, or extract data (unapply), and in this case you'd be doing neither (arity ≠ patterns in Scala). You can avoid the double set of parens a different way. Like this:

```scala
    ReactComponentB[List[String]]("app")
      .initialState(State(0))
      .backend(new Backend(_))
      .render { (P, S, B) ⇒
        val a = S.index
        div(s"State: $a")
      }.build
```

How's that?

elacin:
Well yes, that solves it.

the `case` would quickly be needed though, if you for example made a ReactComponentB[(Int, Int)] and wanted to directly unpack the props in the function. I am not sure how many people code like me - I frequently add those `case`s to avoid thinking about when they are needed.

I understand it if APIs are frozen or better kept in line with javascript, but this would have been my proposal anyhow:
```scala
  // ===================================================================================================================
[...]

PR #139, opened 2015-06-07
ListenableBuilder

elacin:
So, i had some problems configuring my components to listen to events.

My main complaint is that `Listenable.install` needs access to the full types of the `ComponentScopeM` and the type parameter of the `Broadcaster[A]`, yet doesn't have a place to get them.

This means that i ended up manually annotating types like this:
```scala
component.configure(Listenable.install(_ => model, ($: ComponentScopeM[Props, State, Backend, TopNode]) => (_: Unit) => $.forceUpdate()))
```

With changes like the ones im proposing, this ends up like this:
```scala
component.listenToFromProps(_.model).ignoringData.handleEvent(_.forceUpdate())
```

I havent looked at all at `EventListener`, it would be even better if we could manage to get a uniform API for listening to dom events as well.

What do you think?

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/ReactComponentB.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/ReactComponentB.scala
@@ -1,5 +1,7 @@
 package japgolly.scalajs.react
 
+import japgolly.scalajs.react.ReactComponentC.ReqProps
+
 import scala.scalajs.js.{Any => JAny, Array => JArray, _}
 import Internal._
 
@@ -10,9 +12,14 @@ object ReactComponentB {
 
   @inline def apply[Props](name: String) = new P[Props](name)
 
-  implicit def defaultDomType[P,S,B](c: PSBN[P,S,B]) = c.domType[TopNode]
-  implicit def defaultProps[P,S,B,N <: TopNode](c: ReactComponentB[P,S,B,N]) = c.propsRequired
-  implicit def defaultDomTypeAndProps[P,S,B](c: PSBN[P,S,B]) = defaultProps(defaultDomType(c))
+  implicit def defaultDomType[P,S,B](c: PSBN[P,S,B]): ReactComponentB[P, S, B, TopNode] =
+    c.domType[TopNode]
+
+  implicit def defaultProps[P,S,B,N <: TopNode](c: ReactComponentB[P,S,B,N]): c.Builder[ReqProps[P, S, B, N]] =
+    c.propsRequired
+
+  implicit def defaultDomTypeAndProps[P,S,B](c: PSBN[P,S,B]): ReactComponentB[P, S, B, TopNode]#Builder[ReqProps[P, S, B, TopNode]] =
+    defaultProps(defaultDomType(c))
 
   // ===================================================================================================================
   // Convenience
--- a/extra/src/main/scala/japgolly/scalajs/react/extra/Listenable.scala
+++ b/extra/src/main/scala/japgolly/scalajs/react/extra/Listenable.scala
@@ -20,7 +20,7 @@ trait Listenable[A] {
   def register(f: A => Unit): () => Unit
 }
 
-object Listenable {
+@deprecated("Use import japgolly.scalajs.react.extra._; component.listenTo() instead", "2015-06-07") object Listenable {
 
   def install[P, S, B <: OnUnmount, N <: TopNode, A](f: P => Listenable[A], g: ComponentScopeM[P, S, B, N] => A => Unit) =
     OnUnmount.install compose ((_: ReactComponentB[P, S, B, N])
@@ -34,5 +34,4 @@ object Listenable {
 
   def installSF[P, S, B <: OnUnmount, N <: TopNode, M[_], A](f: P => Listenable[A], g: A => ReactST[M, S, Unit])(implicit M: M ~> IO, F: ChangeFilter[S]) =
     installIO[P, S, B, N, A](f, (t, a) => t.runStateF(g(a)))
-
 }
--- a/extra/src/main/scala/japgolly/scalajs/react/extra/ListenableB1.scala
+++ b/extra/src/main/scala/japgolly/scalajs/react/extra/ListenableB1.scala
@@ -1,0 +1,67 @@
+package japgolly.scalajs.react.extra
+
+import japgolly.scalajs.react.ScalazReact._
+import japgolly.scalajs.react._
+
+import scalaz._
+import scalaz.effect.IO
+
+final case class ListenableB1[P, S, B <: OnUnmount, N <: TopNode, A]
+                             (    c: ReactComponentB[P, S, B, N])
+                             (val f: P => Listenable[A]) { outer ⇒
+  
+  private type Comp = ReactComponentB[P, S, B, N]
+
+  private[this] def handleEvent(g: (ComponentScopeM[P, S, B, N], A) => Unit): Comp =
+    c.componentDidMount($ => $.backend onUnmountF f($.props).register(a ⇒ g($, a)))
+     .configure(OnUnmount.install)
+
+  private[this] def handleEventIO(g: (ComponentScopeM[P, S, B, N], A) => IO[Unit]): Comp =
+    handleEvent(($, a) => g($, a).unsafePerformIO())
+
+  sealed class ListenableB2[In](mapIn: ((ComponentScopeM[P, S, B, N], A)) ⇒ In){
+    def handleEvent(g: In => Unit): Comp =
+      outer.handleEvent(($, a) ⇒ g(mapIn(($, a))))
+
+    def handleEventIO(g: In => IO[Unit]): Comp =
+      handleEvent(in => g(in).unsafePerformIO())
+
+    def handleEventS[M[_]](g: In => ReactST[M, S, Unit])(implicit M: M ~> IO): Comp =
+      outer.handleEventIO{case (t, a) => t.runState(g(mapIn(t, a)))}
+
+    def handleEventSF[M[_]](g: In => ReactST[M, S, Unit])(implicit M: M ~> IO, F: ChangeFilter[S]): Comp =
+      outer.handleEventIO{case (t, a) => t.runStateF(g(mapIn(t, a)))}
+  }
+
+  object ignoringProps   extends ListenableB2[A](c ⇒ c._2)
+  object ignoringData    extends ListenableB2[ComponentScopeM[P, S, B, N]](_._1)
+  object ignoringNothing extends ListenableB2[(ComponentScopeM[P, S, B, N], A)](identity)
+
+  object ignoringInput /* ListenableB2[Unit](_ => ()) //can not abstract over by-names, so inlining */ {
+    def handleEvent(g: ⇒ Unit): Comp =
+      outer.handleEvent((_, _) ⇒ g)
+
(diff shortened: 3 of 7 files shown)


japgolly:
Hey @elacin, I appreciate the effort you put in here but I wish you'd asked me first because this isn't a direction I want to go, and I will be sorry if you did nice work only to have it rejected. :(

Firstly, I'm not convinced there's a problem with things the way they are. I've used `Listenable.install` without specifying all the component's types, in fact in this PR where you replace usage you can even see it being used and correctly inferring the types. It can seem like it's not going to infer the types correctly but that is a symptom of Scala's really shit, really stupid half-thought-out inference rules. They didn't do a proper job and then proceeded to add special cases it when they really needed for stdlib. (I really hope they do better with dotty. It's possibly the thing I hate the most about Scala.) Seeing as it *can* and *does* work in my experiences, I'd rather a PR demonstrating and/or documenting pitfalls of how to get it to compile properly without you needing to specify types manually.

Secondly, if there's a flaw that I'm not aware of and the whole `.configure`/mixin-as-endofunction approach is wrong, then I want to fix it as a whole. Treating `Listenable` as a special snowflake means all other mixins suffer from the flaw without being fixed, and that we lose uniformity in the library.

elacin:
Right, don't worry - this was the hacking i did to understand how it worked - at least I managed that :)

Also i figured you should see it, because it highlights again an issue that I have raised another variation of before ( #116 ). It might be that my dialect of scala is a juust different enough from yours so that you don't run into it as much, but i feel I'm constantly fighting the compiler while using this library.

For example talking the compiler into choosing the right overloaded method, switching between () and {} braces, annotating types where i wouldn't expect to (frequently on underscores even) and so on, so i usually end up trying to write wrappers to make my life easier - like this one.

I think it would benefit future users if we made it easier to avoid some of those pitfalls. To that end, this PR is perhaps not as much a proposal for an event API as it is a suggestion for how we could write a an API that is a bit more readable (IMO), and provides good type inference OOTB.

Feel free to close this, but i think it's worth thinking about, again.

elacin:
Also, yes. I didn't mention it directly, but `configure` might best be treated as an implementation detail hidden behind something else. It's great because it's general and you can build pretty much anything on top of it, but Im advocating we give the user a bit more guidance here :)

japgolly:
> i feel I'm constantly fighting the compiler while using this library

I'm surprised and sorry that feel that way. When I use this library I pretty much never have that experience and I'd like to help you and other users not feel that way. The fact I use this library intensively in a way that I don't run into these issues and very rarely have to specify types, leads me to believe that the library itself is fine (hence my hesitation to change it to address I problem that I can't see).

How about we do this: over time as you run into type inference issues can you note them and then send them to me to have a look at? If the solution is a small shift in style, then I'm going to have to just suggest a change in style be the solution. Fighting scalac for style (which is 99% of the time just hyper-subjective personal preference) isn't worth it - I'd rather spend time writing and maintaining code that provides functionality instead of style - better to just be flexible and work with scalac. In these cases we can put a doc together that shows some troublesome styles and how to avoid them.

If there are cases that don't fit the aforementioned criteria (and/or you can convince me they are a serious problem), how about we explore what we could/should change in the lib depending on the case.

> `configure` might best be treated as an implementation detail [...] Im advocating we give the user a bit more guidance here

I must disagree about `configure` being an implementation detail. In fact it's quite the opposite in that we could delete it completely and apart from dsl-flow, we'd lose nothing; it does nothing in terms of implementation. You could do `val b1 = ReactComponentB[…](…)…; val b2 = Listenable.install(…)(b1); val component = b2.build`.
That being said, I'm not dismissing your point that more user guidance would be helpful - I'm open to ideas on that.

elacin:
> I'm surprised and sorry that feel that way. When I use this library I pretty much never have that experience and I'd like to help you and other users not feel that way

Note that I'm not denying that things mostly work fine, in fact this is one of the finest libraries i have used, so thank you for that! That's also why i felt compelled to hack on it for a while to see if i could help make it better :)


> How about we do this: over time as you run into type inference issues can you note them and then send them to me to have a look at?

That seems like a viable strategy going forward. I could indeed try to compile a list of small things i come across, so there would be a bigger picture to discuss. That said:

> Fighting scalac for style (which is 99% of the time just hyper-subjective personal preference) isn't worth it - I'd rather spend time writing and maintaining code that provides functionality instead of style - better to just be flexible and work with scalac. 

I think this is the core of what i want to communicate here. We have both noted that scalac has its imperfections, and that when given a a choice, it will frequently do the wrong thing. I think good API design for a scala library prohibits scalac from doing the wrong thing, and makes the library impossible to use it incorrectly. 

Case in point, in the following snippet:
```scala
component.configure(Listenable.installSF(<f>))
```
scalac will happily infer the correct type for `<f>` once they line up. What doesn't work well is code completion and error messages *before* you have gotten them right. This is exactly the kind of behaviour that i think makes the library a bit difficult to use.

On the other hand, from the PR:
```scala
component
  .listenToFromProps(p => p.broadcaster) //returns ListenableB1[P, S, B, N, A], with A defined by the passed in function
  .ignoringProps //casts the expected function parameter from (P, A) to (A)
  .handleEvent(handleA) //takes a function from A => Unit
```
This threads through and partially applies types and parameters, so all of scalac, intellij and the user will have a clear idea of types at hand. This is *difficult* to use incorrectly, and the partial application of types guides the user naturally through using it. Indeed it's built along the same line as `ReactComponentB`.

But yeah, I'll come up with some more examples :)

japgolly:
> this is one of the finest libraries i have used, so thank you for that!

Wow that's wonderful to hear! Thank you. If you lived in Sydney I'd buy you a :beer: .

> I think good API design for a scala library prohibits scalac from doing the wrong thing, and makes the library impossible to use it incorrectly.

I see what you mean, and I agree! But there's one small point on which we might disagree. I would say that scalac not compiling means that the API is not letting you do the wrong thing - that's why the errors flow so easily :D But look, I do take your point. I just don't 100% agree on the proposed solution right now. The main reasons are that it needs to be solved for each mixin *implementation* rather than once at the interface, and also I don't want to go down the hole of predicting what will be needed, providing a DSL there only to find someone needs something different 3 days later.

The best solution I think would be some kind of generic way that applied to all. Maybe even a rethink of how we can do mixins better. Maybe some tricks with implicits and type variance (I have an idea I'm hesitant to even explore right now). I'll leave the door open on this.

japgolly:
Oh and I'll close this PR too.

elacin:
hehe, i'll take you up on that beer if im ever around.

Anyways, it occurred to me that a much less intrusive change like this
```diff
-  def install[P, S, B <: OnUnmount, N <: TopNode, A](f: P => Listenable[A], g: ComponentScopeM[P, S, B, N] => A => Unit) =
+  def install[P, S, B <: OnUnmount, N <: TopNode, A](f: P => Listenable[A])(g: ComponentScopeM[P, S, B, N] => A => Unit) =
```
would also go a long way towards fixing the type inference problem. Anyways, I'll leave it at that :)

japgolly:
> hehe, i'll take you up on that beer if im ever around.

Do!

> Anyways, it occurred to me that a much less intrusive change like this

I don't think that would work because Scala would then choke trying to infer the `P` in the first argument. Scala's type inference problems are such a pain in the ass. I was thinking about this a bit more, I feel like we should be able to make a quick DSL for a forall-like function with DSL to just set constraints.

Something like this:
```scala
  /** Mixin type DSL */
  trait Mixin[
      PLB, PUB >: PLB,
      SLB, SUB >: SLB,
      BLB, BUB >: BLB,
      NLB <: TopNode, NUB >: NLB <: TopNode] {
    def plb[T <: PLB]: Mixin[T, PUB, SLB, SUB, BLB, BUB, NLB, NUB] = ???
    def slb[T <: SLB]: Mixin[PLB, PUB, T, SUB, BLB, BUB, NLB, NUB] = ???
    def blb[T <: BLB]: Mixin[PLB, PUB, SLB, SUB, T, BUB, NLB, NUB] = ???
    def s[T >: SLB <: SUB]: Mixin[PLB, PUB, T, T, BLB, BUB, NLB, NUB] = ???

      def action[
        P >: PLB <: PUB,
        S >: SLB <: SUB,
        B >: BLB <: BUB,
        N >: NLB <: NUB
        ](fn: ReactComponentB[P, S, B, N] => ReactComponentB[P, S, B, N]) =

      new Mixin2[
        PLB, PUB,
        SLB, SUB,
        BLB, BUB,
        NLB, NUB] {
        override def apply[
          P >: PLB <: PUB,
          S >: SLB <: SUB,
          B >: BLB <: BUB,
          N >: NLB <: NUB
          ](c: ReactComponentB[P, S, B, N]): ReactComponentB[P, S, B, N] =
          fn(c) // this wont work, der!

      }
  }

  /** Mixin ready to be used */
  trait Mixin2[
    PLB, PUB >: PLB,
    SLB, SUB >: SLB,
    BLB, BUB >: BLB,
    NLB <: TopNode, NUB >: NLB <: TopNode] {

    def apply[
      P >: PLB <: PUB,
      S >: SLB <: SUB,
      B >: BLB <: BUB,
      N >: NLB <: NUB
      ](c: ReactComponentB[P, S, B, N]): ReactComponentB[P, S, B, N]
  }

  def empty: Mixin[Any, Nothing, Any, Nothing, Any, Nothing, TopNode, Nothing] = ...
```

and then usage would be like this:
```scala
// Applies to ReactComponentB[?, Int, ? <: OnUnmount, ?]
  val mixin = empty.s[Int].blb[OnUnmount].action(
    _.componentDidMount($ => $.backend.onUnmount {
      println($.state min 0)
    }))
```

But it sort of obviously doesn't work. I wonder if there's a way though....

japgolly:
Maybe a macro would help? If you can call `.action` it's proof that you can supply a function for all types within bounds, which is what we want to store without actually applying the bounds.

japgolly:
Actually we *do* have a proof. We just can't inform the compiler of it. We could do this:
```scala
      def action
        [P >: PLB <: PUB, S >: SLB <: SUB, B >: BLB <: BUB, N >: NLB <: NUB]
        (fn: ReactComponentB[P, S, B, N] => ReactComponentB[P, S, B, N]) =

      new Mixin2[PLB, PUB, SLB, SUB, BLB, BUB, NLB, NUB] {
        override def apply
            [P2 >: PLB <: PUB, S2 >: SLB <: SUB, B2 >: BLB <: BUB, N2 >: NLB <: NUB]
            (c: ReactComponentB[P2, S2, B2, N2]): ReactComponentB[P2, S2, B2, N2] =
          // well, we have a proof, we just can inform the compiler of it
          fn.asInstanceOf[ReactComponentB[P2, S2, B2, N2] => ReactComponentB[P2, S2, B2, N2]](c)
      }
```

japgolly:
Holy crap! This would work and would remove a lot of type inference problems and boilerplate. Opening an issue...

elacin:
Holy crap indeed! I would have to meditate a bit on it to properly get what youre doing there, will try to play a bit with it later today to see if i can give any feedback :)

> I don't think that would work because Scala would then choke trying to infer the P in the first argument. Scala's type inference problems are such a pain in the ass. I was thinking about this a bit more, I feel like we should be able to make a quick DSL for a forall-like function with DSL to just set constraints.

Yeah, i guess you're right. Would need to be three parameter sets to properly guide type inferring.

Issue #132, opened 2015-05-30
Implement ReactJS context


slorber:
Contexts can be used to pass data to a whole hierarchy of components.
https://www.tildedave.com/2014/11/15/introduction-to-contexts-in-react-js.html
(this may be outdated but gives you the idea)

This is undocumented by React on purpose as the feature may evolve  (it changed a bit in 0.13 I think) but it won't disappear and is already used by a lot of frameworks (like Flux frameworks, React-Intl etc...)

I use this to put data that barely never change like current user id, or app language.

Typically I think things like your broadcast/listenable feature should be injected in an app by a context as it removes the need for a singleton eventbus. One should be able to broadcast messages from outside of React components, as well a one should be able to listen from outside (I do this for example to handle communication with iframes or websockets: don't want to put this in a component)

chandu0101:
looks like its a must feature we should implement! .

from https://github.com/wav/material-ui-scalajs-react : 
>scalajs-react currently doesn't have a way to define a react context. A fork is used as a hack.

here is the hack he is using 

https://github.com/wav/scalajs-react/commit/0510e35ada18f2bd8f5065572c261086bad30266

i wonder why he didn't sent a PR!

wav:
@chandu0101. I was planning to tonight or tomorrow sometime time when I've fleshed it out a little more. Currently it was a hack to see if it works. Ideally, you would pass in a trait that had the two context properties and the builder would add it to the spec. I haven't had any issues with it thus far.

chandu0101:
@wav cool  ok , waiting for PR  . I think it also useful to have a unsafespec method, in future if new feature comes we can atleast get it working in un safe manner instead of waiting for new release!

offtopic : any specific reason for disabling issues in your material-ui wrapper project ?

japgolly:
Just briefly for now: could you please include a small test case in the PR.
I want to confirm it works now on React 0.12, but will also need to confirm
it doesn't break when we upgrade to 0.14. Thanks!

wav:
Ok will do tonight. At work atm.

japgolly:
Take your time, there's no rush :)
Next release is probably about 2-3 weeks away.

slorber:
Notice in 0.14 there's the official switch from owner-context to parent-context.

If you use "portals" for exemple for modals/tooltips you have to use a new method `renderSubtreeIntoContainer`

See 
https://github.com/facebook/react/issues/4081
https://github.com/rackt/react-modal/issues/37


[...]

PR #143, opened 2015-06-13
Unsfasespec

chandu0101:
addresses #132 

actually code is from @wav fork , looks like hes busy so just want to take a chance to help .. 

apart from #132 concerns ,its always better to have unsafe method (to allow end user what ever the hell he want to do ) , so that in future users don't need to wait for next version releases .


japgolly:
Cool. Can you do the following:

1. Rename to `configureSpec`.
2. Align those colons on line 79.
3. On line 196, move the application of the function down. It should be the last thing called so that the fn receives the spec fully built instead of half built.
4. Squash ye commits.

chandu0101:
i have no idea whats going on here ,let me know if its okay ?

japgolly:
I got it: 5a76e382f03ca2c8124931112ed5e0c1878daf89

chandu0101:
aww , thanks a lot mate :+1:

PR #144, opened 2015-06-14
Add the ability to define and use an untyped react context.

wav:
Tests ok on 0.12.1 and 0.13.3.

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/React.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/React.scala
@@ -220,6 +220,7 @@ trait ComponentScopeU[Props, State, +Backend]
 /** Type of an unmounted component's `this` as is available to `getChildContext()`. */
 trait ComponentScopeCU[Props]
   extends ComponentScope_C
+  with ComponentScope_P[Props]
 
 /** Type of a component's `this` scope during componentWillUpdate. */
 trait ComponentScopeWU[Props, +State, +Backend, +Node <: TopNode]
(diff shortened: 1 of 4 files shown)


chandu0101:
wow ,aweomse dude :) :+1:

wav:
I should add a test case for accessing props. Awaiting review.

@chandu0101 Cheers.

omidb:
:+1:

japgolly:
This looks very good. I expect to get a good full day in for OSS later in the week and I'll review properly then. For now, just a small thing: could you please change `Option` to `js.UndefOr`? It just saves a tiny bit of work & mem.

wav:
Should probably add another 'derived' case that doesn't allow access to the parent context, just props.

japgolly:
Ok I've just had an 80% read through #132, and its links, and this PR. Honestly I'm not comfortable with this. You've put in good work and I don't mind supporting an undocumented React feature but examining this PR in more detail I see that this goes against all the type-safety and benefits of using Scala over JavaScript.

Why is this desired? Is it purely for JS interop or is there another reason?

I can see that there are at least 3 people that want this feature so we'll try to it done in a way that makes everyone happy. Maybe this needs a big change for full type safety, or maybe we need less type safety and just a few points of access such that using it means you're basically in JS-land.

wav:
I did expect that this PR it would go against the type-safey mantra unless a big/breaking change is made.

My desire is JS interop, #143 will do me fine. I'm happy to close this.

chandu0101:
@wav could you please send PR for #143  , my git is messed up. please follow @japgolly instructions mentioned in ticket.

Thank You
Chandra

japgolly:
Don't worry it's in: 5a76e382f03ca2c8124931112ed5e0c1878daf89

@wav So this is a bit middle-of-the-road between Scala & JS. If you come up with a way that's extremely  JS-only and/or a way that's extremely Scala with strong type-safety, please feel free to reopen. Cheers

PR #146, opened 2015-06-16
change return type of isMounted()

chandu0101:
i just got this error 

```scala
Uncaught scala.scalajs.runtime.UndefinedBehaviorError: An undefined behavior was detected: undefined is not an instance of java.lang.Boolean
```

when calling isMounted()

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/React.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/React.scala
@@ -164,7 +164,7 @@ trait ReactComponentM[Props, State, +Backend, +Node <: TopNode]
 
 /** Methods always available. */
 trait ComponentScope_A extends Object {
-  def isMounted(): Boolean = js.native
+  def isMounted(): UndefOr[Boolean] = js.native
 }
 
 trait ComponentScope_P[+Props] extends Object {


japgolly:
What did you get it on? Did you manually cast anything before getting the result?

chandu0101:
no i am just doing if(node.isMounted() )

japgolly:
node comes from where?

chandu0101:
selectedNode : BackendScope[NodeProps,NodeState]

i am house keeping scopes in my component when ever there is a change! ,i have to dig more to se what exactly i am doing , i wrote this code 6 months back :) 

nevertheless any issues if we have UndefOr[Boolean] ?  and btw i am using react 0.13.3

japgolly:
I don't want to fix this before we understand exactly what fix needs to be, we should recreate it in a test first. Maybe fix is to remove a trait so `isMounted` isn't available, or even rename & hide & add facade. Too soon to tell. But I don't want `isMounted` return type to have 3 possible values - should be 2.

chandu0101:
that makes sense , let me see if i can reproduce with a test case :)

chandu0101:
as i suspected this is happening in reactjs 0.13.x

here is test case 

```scala
 var cache : BackendScope[_,_] = null
      class Backend(t : BackendScope[_,_]) {
        def cacheScope() = cache = t
      }
      val C = ReactComponentB[Unit]("ChildComp")
        .stateless
        .backend(new Backend(_))
        .render(P => div())
        .componentWillMount(scope => assert(!scope.isMounted()))
        .componentDidMount(scope => scope.backend.cacheScope())
        .buildU

      case class State(toggle : Boolean = true)

      val P = ReactComponentB[Unit]("ParentComp")
        .initialState(State())
        .noBackend
        .render((P,S,B) =>
          div(
            S.toggle ?= C()
          )
        )
        .componentDidMount(scope => {
        assert(cache.isMounted()) // test cached scope also
        scope.modState(s => s.copy(toggle = !s.toggle))
      })
        .buildU

      val instance =  ReactTestUtils.renderIntoDocument(P())

      assert(!cache.isMounted()) // check after child mounted

```

japgolly:
This ↓ can never return `undefined` right?
```js
return (
  internalInstance &&
  internalInstance !== ReactLifeCycle.currentlyMountingInstance
);
```

The comment on that fn seems to support my supposition that I'm not crazy and JS isn't *that* crazy. How can that function above result in `undefined is not an instance of java.lang.Boolean` for us?

chandu0101:
u r not , but JS is :p 

```js
// try thisin console
undefined&&true
= undefined in console 
```

japgolly:
Seriously wtf is wrong with JS???
```js
> undefined && true
undefined

> undefined && false
undefined

> true && undefined
undefined

> false && undefined
false
```

Awesome that you tracked this down. It looks like a React bug to me. If you you raise over there they'll probably fix it pretty quickly.

chandu0101:
ah its fixed already https://github.com/facebook/react/issues/3814 

will be in next version of react :)

japgolly:
excellent

PR #154, opened 2015-07-02
Shouldn't these be elements instead of attributes?

i-am-the-slime:


japgolly:
They sure should. Thanks!

Issue #157, opened 2015-07-12
Support for JS Component Types


imcharsi:
Hi.

At first, My English is poor. I'm sorry about that.

I think that it would be better if we can use below method.
```
createElement(component: js.Dynamic, props: js.Object, children: ReactNode*): ReactComponentU_
```

In javascript, we can write below code with Material-UI.
```
React.createElement(mui.TextField, {someProp...}, someChildren...);
```

[Here is chandu0101's example.](https://github.com/chandu0101/scalajs-react-components/blob/master/core/src/main/scala/chandu0101/scalajs/react/components/materialui/package.scala#L42)
```
    val TextField: js.Dynamic = js.native
```

And, [Here.](https://github.com/chandu0101/scalajs-react-components/blob/master/core/src/main/scala/chandu0101/scalajs/react/components/materialui/MuiTextField.scala#L69)
```
    val f = React.asInstanceOf[js.Dynamic].createFactory(Mui.TextField)
    f(p).asInstanceOf[ReactComponentU_]
```

So, If we can use above method, we can write below code.
```
React.createElement(Mui.TextField, p)
```

Thanks.

chandu0101:
+1

japgolly:
Using `js.Dynamic` is lazy. Should be using JS types like `ReactComponentC_`, `ReactComponentU_`, `ReactComponentM_` for JS components. Is there a type missing for what gets fed into `createElement`?

With Scala, where … = `ReactComponent`, it's
`…B` ⇒  `…Spec` ⇒ `…Type` ⇒ `…C` ⇒ `…U` ⇒ `…M`.

I don't mind a more JS-friendly `createElement` being added but first you'll need to replace `js.Dynamic` with something more appropriate.

(I notice that Scala components' internal JS component ctor is typed `ReactComponentCU`. That seems a bit weird - either I'm reading too quickly now or I was accidentally inconsistent in the past.  Either way I wonder if that type might overlap with what you need here.)

imcharsi:
Hi.

Thanks for reply.

At first, If below my suggestion is not your intended usage, please tell me about your example usage.

So, [with below API](https://github.com/japgolly/scalajs-react/blob/master/core/src/main/scala/japgolly/scalajs/react/React.scala#L18),

```
  def createFactory[P,S,B,N <: TopNode](t: ReactComponentType[P,S,B,N]): ReactComponentCU[P,S,B,N] = js.native
```

I tried below code.

```
[...]

PR #158, opened 2015-07-14
Support for JS Component Types.

imcharsi:
Hi.

As we talked at https://github.com/japgolly/scalajs-react/issues/157, I tried this pull request.

Review please.


japgolly:
Hmmm interesting... I'm very happy that you supplied a proper test case for this :) Nice work.

@chandu0101 You seem to do a lot of work with JS React components. Do  you have any thoughts on this?

chandu0101:
on a quick glance ,so end user have to write this code
```scala
React.createFactory(SampleReactComponent)(SampleReactComponentProperty(ref = ref, propOne = "123"))
```
to use custom component SampleReactComponent ? 
if its the case sorry i'll disagree with you  , for me the current facade way that we're using
https://github.com/japgolly/scalajs-react/blob/master/core/src/main/scala/japgolly/scalajs/react/Addons.scala#L13
is more elegant , because end user using it like an API ,he'll pass inputs required (no need to worry about  props ,createFactory... )

imcharsi:
Hi.

Thanks for your replies.

I understand that the API is something inconvenient, too.

For `React.createFactory`, we can write some codes like below with suggested API.

```
object Xxx {
  val Xxx = React.createFactory(SomeComponent)
}

import Xxx.Xxx

Xxx(SomeProperty(...))
```

By the way, with Scala-React-API, we use the API like below.

```
object Xxx {
  val Xxx = ReactComponentB[SomeProperty]("S")....build
}

import Xxx.Xxx

Xxx(SomeProperty(...))
```

At least, for API usage I don't think that there are any significantly different.

chandu0101:
>At least, for API usage I don't think that there are any significantly different
agree , 
currently i am creating extra apply methods to overcome this for scala components

```scala
object GeoJsonFeatureFarm {

  val component = ReactComponentB[Props](..)
  case class Props(geojsonFeature: Option[GeoJsonFeature],index :Int, onAddUpdateSuccess : (GeoJsonFeature,Int) => Unit ,onCancel :() => Unit)

  def apply(geojsonFeature: Option[GeoJsonFeature],index : Int,onAddUpdateSuccess : (GeoJsonFeature,Int) => Unit ,onCancel : () => Unit = null , ref: js.UndefOr[String] = "", key: js.Any = {}) = component.set(key, ref)(Props(geojsonFeature,index,onAddUpdateSuccess,onCancel))

}
```
to get uniform styles all for all components!

my point is we don't need to follow existing scalajs-react API style 

in next version of scalajs-react ( that supports reactjs 0.14) will have a new API kind of similar to existing facade way.

>japgolly - In the next major release I'll probably lump all the big changing things together (components as classes, React 14, unify -->/==>/~~>)

that being said ignore me if you  prefer u r js world api , its always good to have more options people choose what they love :)

imcharsi:
Hi.

Thanks for reply.

> my point is we don't need to follow existing scalajs-react API style.

Right.

> Its always good to have more options, people choose what they like :)

Yes. I hope so.

japgolly:
@imcharsi Sorry to keep you waiting. Could you please make some changes:

1. The `Props` type in all the new JS traits should have an upper-bound of `js.Any`. Example, please replace `JSComponentType[Props,` with `JSComponentType[Props <: js.Any,`.
2. Please replace `JS` with `Js`, so `JSComponentType` becomes `JsComponentType`.
3. Could you please write a doc that:
  1. Summarises the types like in [TYPES.md](https://github.com/japgolly/scalajs-react/blob/master/doc/TYPES.md#components)
  2. Shows how to write a facade for, and use a JS component.
4. The JS types have `Props` but no `State`. They should have `State` too, no?

imcharsi:
Hi.

I'm sorry for lately reply.

I'll try as you said.

Thanks.

imcharsi:
Hi.

I tried as you said.

But, you may need to check correction about spell and grammar especially. I'm sorry about this.

Review please.

japgolly:
@imcharsi I'm not going to give this a full review today because I have other work I need to get done, so there will probably be another delay. I want to quickly say, I had a skim just now (a quick look), and it looks great! So thank you very much! I'll get back to you soon.

japgolly:
Hi @imcharsi. I just had a proper read though. It looks good to me, I'll merge it now. Big thank you for your effort and patience. I especially appeciate the doc and tests. :+1: :+1:

japgolly:
@imcharsi FYI I made a few follow-up edits ↓, nearly all just tweaking the documentation. Thanks again.

https://github.com/japgolly/scalajs-react/compare/2b74e842...1f6de8d

PR #162, opened 2015-07-30
Suggest about monkey patching.

imcharsi:
Hi.

I want to suggest about monkey patching. See [here](http://www.scala-js.org/doc/calling-javascript.html).

Instead of

```
object React extends js.Object {
  ...
}

object ReactTestUtils extends js.Object {
  ...
}

object Simulate extends js.Object {
  ...
}
```

How do you think about like below?

```
trait React extends js.Object {
  ...
}

object React extends React

trait ReactTestUtils extends js.Object {
  ...
}

object ReactTestUtils extends ReactTestUtils

trait Simulate extends js.Object {
  ...
}

object Simulate extends Simulate

```

If those objects are changed like that, we can these as alternative for something like below.

```
object SomethingCustomConverter {
  implicit def conv(react:React) = react.asInstanceOf[WrapperFindDOMNode]
}

trait WrapperFindDOMNode extends React {
  def findDOMNode[N <: TopNode](component:ReactComponentM_):N = js.native
}

import SomethingCustomConverter._

React.findDOMNode(someComponent)
```

or like below
```
object TouchTapConverter {
  implicit def conv(simulate:Simulate) = simulate.asInstaceOf[WrapperTouchTap]
}

trait WrapperTouchTap extends Simulate {
  def touchTap(...) = js.native
}

import TouchTapConverter._

Simulation(ReactTestUtils.Simulate.touchTap(_))
```

Review please.


japgolly:
No problem.
Merged in 3aba80c6b4560be5a0308f15e08023b87f26c045.
Thanks @imcharsi

PR #179, opened 2015-09-04
Add reusability with precision + js.Date/java.util.Date/UUID

matthughes:
Float/Double withPrecision plus js.Date/java.util.Date/UUID cause I use those types all the time.  Up to you though.

### DIFF
--- a/extra/src/main/scala/japgolly/scalajs/react/extra/Reusability.scala
+++ b/extra/src/main/scala/japgolly/scalajs/react/extra/Reusability.scala
@@ -1,5 +1,8 @@
 package japgolly.scalajs.react.extra
 
+import java.util.{Date, UUID}
+import scala.scalajs.js.{ Date => JsDate }
+
 import scalaz._
 import scalaz.effect.IO
 import japgolly.scalajs.react.{ComponentScopeM, ReactComponentB, TopNode, ReactElement}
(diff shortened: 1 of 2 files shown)


extra/src/main/scala/japgolly/scalajs/react/extra/Reusability.scala
```diff
@@ -77,16 +80,23 @@ object Reusability {
...
+  @inline implicit def reusableChar       : Reusability[Char   ] = by_==
+  @inline implicit def reusableShort      : Reusability[Short  ] = by_==
+  @inline implicit def reusableInt        : Reusability[Int    ] = by_==
+  @inline implicit def reusableLong       : Reusability[Long   ] = by_==
+  @inline implicit def reusableString     : Reusability[String ] = by_==
+  @inline implicit def reusableJavaDate   : Reusability[Date   ] = by_==
+  @inline implicit def reusableJsDate     : Reusability[JsDate ] = by_==
+  @inline implicit def reusableUUID       : Reusability[UUID   ] = by_==
+
+  implicit def doubleWithPrecision(precision: Double): Reusability[Double] = fn((x, y) =>
```
japgolly: Can't be `implicit` here - it needs an arg.
Can we rename? How about just `def double(tolerance: Double): Reusability[Double]`
And same with float

japgolly: Oh and could you just move it up above the `// Instances` line?
By `// Instances` I meant "implicit, default instances"

japgolly:
Otherwise LGTM

japgolly:
Oh, are you sure that univ equality holds for `Date` and `js.Date`? Could you chuck in a quick unit test?

matthughes:
Updated.  Bah, I should have known js.Date wouldn't doesn't equality.  Made other requested changes.

japgolly:
Oh lucky. That turned out well! Thanks

PR #183, opened 2015-09-12
More specific jsDependency

alexmojaki:
When I tried using just "react.js" this in a Play project I got this error:
JSLibResolveException: Some references to JS libraries could not be resolved:
- Ambiguous reference to a JS library: react.js
Possible paths found on the classpath:
- META-INF/resources/webjars/react/0.12.2/vendor/fbtransform/transforms/react.js
- META-INF/resources/webjars/react/0.12.2/react.js
originating from: client:compile
Specifying the more specific path solved it.


japgolly:
I've had a similar problem with a different library but not React. Weird.

PR #186, opened 2015-09-15
Consolidated SetInterval and SetTimer support.

matthughes:


extra/src/main/scala/japgolly/scalajs/react/extra/TimerSupport.scala
```diff
@@ -0,0 +1,52 @@
...
+  }
+
+  /** Provides `setInterval`-like behavior insuring that the time between calls of `f` is *at least* the `timeout`. */
+  final def setGuaranteedInterval(f: Callback, interval: FiniteDuration): Callback= {
+    CallbackTo {
+      val i = RawTimers.setTimeout(() => {
+        f.runNow()
+        setGuaranteedInterval(f, interval)
+        ()
+      }, interval.toMillis)
```
japgolly: Few things. 1) This won't work. Did you put that `()` there to avoid a compiler-error? It was a valid. error in that `setGuaranteedInterval()` is created but never invoked. 2) Best not to call `.runNow()` yourself. 3) What should happen if the callback fails? We should consider that wrt setting the next interval. 4) Should use the new `setTimeout` method, no?

I think something like this is better:

```scala
    val repeat = Callback byName setGuaranteedInterval(f, interval)
    val runSafe = f.attempt.map(_.left.foreach(_.printStackTrace()))
    setTimeout(runSafe >> repeat, interval)
```

japgolly:
Few more things:

1. Could you pls do the rename to `TimerSupport` in a single commit, then add your new features in a second commit?
2. Just like how there's `setInterval` and `setInternalMs` could you please add `...Ms` versions of the two new functions you're adding?
3. Make a quick mention of it in `doc/CHANGELOG-0.10.md`.

That should do it.

matthughes:
k, done.

japgolly:
Looks good but I think you missed [my other comment](https://github.com/japgolly/scalajs-react/pull/186#discussion-diff-39467546) - I think the impl of `setGuaranteedInterval` is still broken. Actually I'll fix it and add another method to `Callback` for a `finally` clause.

japgolly:
FYI this is the bugfix for `setGuaranteedInterval`: 406e26da70257201fb418d728504f4ef16214231

Update to that, yeah, if you're using SNAPSHOT.

matthughes:
Thanks.  I totally missed that comment.

japgolly:
no worries!

FYI I made another change:  af702f9fd4961251ef6600c616f5ad173746eb55
Unlike `setInterval`, `setTimeout` handles needn't be cleared once they schedule up.

PR #195, opened 2015-10-11
withFilter for CallbackOption to enable if in for comprehensions

elacin:
Hah, i was coing to contribute a `zip` for `Callback` and `CallbackOption` - seems you beat me to it. That leaves this:

`withFilter` is to use ifs for `CallbackOption` in for comprehensions without a warning (breaks my build since im normally using most of https://tpolecat.github.io/2014/04/11/scalac-flags.html )


japgolly:
Nice :+1:

PR #196, opened 2015-10-11
A few more events were added at some point

elacin:
Hehe, we've been here before. `onMouseEnter` and `onMouseLeave` are the ones i really want, the rest are just to complete the list i guess


japgolly:
Oh wow, how'd we miss all of that?! Love your work mate :D

PR #197, opened 2015-10-12
Link to todomvc.com

elacin:
Finally got my PR merged at todomvc after quite a while. will update it to 0.10 too, so i figured it's a more useful link


japgolly:
Ah just before v0.10 is released too

Issue #199, opened 2015-10-12
Add DOM types to ReactTags?


japgolly:
@elacin said:

That same commit (https://github.com/elacin/scalajs-react/commit/fb6d0323b655ae80c3f005e161b3d164ff90544f) also shows an experiment with typing `ReactTag`s (so i can haz some type-safety for my `HTMLInputElement => Unit` ref updates), would that be something you're interested in?

About typed `ReactTag`s, I can probably save you the trouble. it comes down to this:
```scala
//covariant, to let a ReactTagOf[html.Paragraph] also be a ReactTagOf[TopNode]
abstract class ReactTagOf[+N <: TopNode] extends DomFrag 
//regain invariance to make compiler happy
private [vdom] final case class ReactTagImpl[N <: TopNode] 
//make sure existing code should be allright
final type ReactTag = japgolly.scalajs.react.vdom.ReactTagOf[TopNode]

// make all tags in HtmlTags.scala gain more specified types
implicit class STStringExt(private val s: String) extends AnyVal {
  def tag[N <: TopNode](implicit namespaceConfig: Namespace): ReactTagOf[N] = ...
  ...
}
```

japgolly:
I think this would be a great idea but it's not high on the list of things I would implement myself. I was thinking about doing this a while back but guessed that it would change too many existing types and decided against it. I quite liked the idea of the component's DOM type being automatically inferred by the result of the render function but then thought that would be too disruptive.

If you delivered it in one of your awesome PRs it would be very welcome! So long as it doesn't add much burden to users/callsites. :)

elacin:
Cool. I'll see if i can find the time for that tonight :)

japgolly:
Closed via #202

PR #202, opened 2015-10-13
More specific types for ReactTag

elacin:
PR for #199

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/vdom/package.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/vdom/package.scala
@@ -19,8 +19,10 @@ package object vdom {
 
   // If you're wondering why abstract class instead of trait, https://issues.scala-lang.org/browse/SI-4767
   abstract class Base extends Implicits {
-    final type ReactTag = japgolly.scalajs.react.vdom.ReactTagOf[TopNode]
-    final type TagMod   = japgolly.scalajs.react.vdom.TagMod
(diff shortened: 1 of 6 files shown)


elacin:
The `RouterCtl` change is not needed, but i figured it wouldnt hurt to do the same there too. Didn't have to pull that invariant trick since this change doesnt include the ref changes i proposed

core/src/main/scala/japgolly/scalajs/react/vdom/package.scala
```diff
@@ -19,7 +19,7 @@ package object vdom {
 
   // If you're wondering why abstract class instead of trait, https://issues.scala-lang.org/browse/SI-4767
   abstract class Base extends Implicits {
-    final type ReactTag = japgolly.scalajs.react.vdom.ReactTag
+    final type ReactTag = japgolly.scalajs.react.vdom.ReactTagOf[TopNode]
```
japgolly: We should probably add a new type alias here for `ReactTagOf`

japgolly:
Only one small comment above. Otherwise, looks good. Would you also mind just making a short note of this towards the bottom of the changelog?

core/src/main/scala/japgolly/scalajs/react/vdom/package.scala
elacin: Ah yes, of course.

elacin:
Will fix both things today, just a day of work first :)

japgolly:
Looks great. Love your work as usual. In it goes; thanks!

PR #203, opened 2015-10-16
Scala.js v 0.6.5 update + ES6 Class support

xsistens:
Hey David,
this is an approach to support ES6 class components within scalajs-react. To accomplish that i had to update the scalajs version to 0.6.5, i think that's a nice side benefit :)

As you can see i have not touched your existing classes or structure at all. So everything that was working before and every further development should be the same as without this addition. The benefit from that solution is, that you can create react components in both ways and share them  between each other. So people that want to create components in the new way, can use all the already existing components made with your work. 

I also included a very very simple example (/es6classes) in the ghpages examples, to show you that you can use all old components in the new ones. 

This change enables other developers to use the ES6 class feature to enhance the project with things like relay for example and still benefit from your work with the router or the dom. 

Maybe you can give that PR a shot ,i would love to see that implemented into your work. If you have any questions feel free to contact me on your gitter room, i'll always check that.

Thanks for your amazing work, i appreciate it a lot!


japgolly:
Hi @xsistens, thanks for this PR! I'm on my way out to breakfast so I'll be brief for now:

* Firstly, *awesome* and thank you. :+1: I can't cover everyone's needs so I'm very happy to see a community contribution like this :D
* We'll need to make some further changes before we merge this. v0.10 has been waiting for ages so I'm going to stick to my (internal) plan and publish v0.10 final later today or tomorrow if I run out of time. We can work on this PR and hopefully get it into the subsequent version, which needn't take too long.
* Is this code for JS components or Scala components? I doubt it can be both and I'll proceed under the assumption that it's for Scala components when I write the following points:
* We'll need to have unit tests that confirm the following:
  * All Scala types work as props/state. I understand React needs props/state to be a `js.Object` which means I think boxing is necessary (see `WrapObj`).
  * It works as expected when nested in another component's render function. (i.e. conversion to `ReactNode` and handling by dom builder & React).
  * It integrates with the rest of the library. I don't want it to seem like a separate, embedded library, I need all the stuff in `extra` that works with existing components to work with this too. 
* We should add another type param for the node type like everything else so `findDOMNode` is type-safe (`+N <: TopNode`). If that makes usage annoying, we can also add a thin layer to make it a default so it doesn't trouble users when they don't care about it.
* We can worry about this at the last step, but documentation will also need to be updated.
* *More maybe?*

That's all that comes to mind now. List looks a bit large but I think it's not as big as it sounds. We can do it! :)

I've got to run, talk soon and thanks again!

ngbinh:
this is so cool! Thanks

japgolly:
Also, if this is for JS, @imcharsi add JS types in 85390382259d2a3a57211b139eec498ce32bcd8f / #158 . We'd need to make sure we don't overlap with / duplicate that work.

xsistens:
Hey,
first thanks for looking into it. I'm busy this morning for about 10 hours from now, when iam back i will try to answer all your questions. What i can say now is that this is for scala components and is different from the work of @imcharsi but i will go into detail later today.

xsistens:
> Firstly, awesome and thank you. :+1: I can't cover everyone's needs so I'm very happy to see a community contribution like this :D

I think thats what open source projects are about. Share knowledge and man power to make products as good as possible and put all the work on the back of the community instead of just a few. So I'am happy that you see it the same way.

> Is this code for JS components or Scala components? I doubt it can be both and I'll proceed under the assumption that it's for Scala components when I write the following points

This PR allows you to write fresh React components within scala.js. So its not about wrapping an already existing JS component. 

> We'll need to have unit tests that confirm the following:

You are totally right about the lack of tests. This is my first real contribution to scala.js and I'm not super familiar with the way testing should be done here. I'll have to investigate into the procedure and have to gather some knowledge of it. Maybe you or the community can help with such regards. 

> All Scala types work as props/state. I understand React needs props/state to be a js.Object which means I think boxing is necessary (see WrapObj).

At the moment it should be able to use all the supported Types of scala.js in the Props or State. It's working similar as in ```ReactComponentB```:

```
@js.native
trait JSProps[P] extends js.Object {
  def key: UndefOr[String] = js.native
  def ref: UndefOr[js.Function] = js.native
  def sprops: P = js.native
  def children: PropsChildren = js.native
}

object JSProps {
  def apply[P](key: UndefOr[String] = undefined,
               ref: UndefOr[js.Function] = undefined,
               sprops: P) = {
    val dict = js.Dictionary[Any]("sprops" -> sprops)
    key.foreach(v => dict.update("key", v))
    ref.foreach(v => dict.update("ref", v))
    dict.asInstanceOf[JSProps[P]]
  }
}
```
```P``` is the Props parameter that is passed to the component and is the return type if you access the Props within the component. You dont have to handle any ```JSProps``` at all, only ```P```. ```JSProps``` is just the wrapper similar to your ```WrapObj``` 

> It works as expected when nested in another component's render function. (i.e. conversion to ReactNode and handling by dom builder & React).

That's definitely what i want to achieve! I think that's a point i can try to test with my actual knowledge about scalajs-react. 

> It integrates with the rest of the library. I don't want it to seem like a separate, embedded library, I need all the stuff in extra that works with existing components to work with this too.

For me that is actually the hardest part, because I have to catch up a lot of the internal knowledge about scalajs-react to see if and how this is possible. I hope I can get some help from you or other people that are very familiar with all these features in the *extra* package. It would be amazing if there is a way to made all this stuff working. I think the *Router* for example is not a problem, but the part about the *Reuseability* iam not sure about yet.

 > We should add another type param for the node type like everything else so findDOMNode is type-safe (+N <: TopNode). If that makes usage annoying, we can also add a thin layer to make it a default so it doesn't trouble users when they don't care about it.

Yeah, i think that's a point that needs more discussion if we want to care about any native part of React. What i mean by that is, that we should ensure that the ```TopNode``` is not restricted to the *DOM* in any point. That's the only way we can ensure that we are able to integrate react-naive in the future.

> We can worry about this at the last step, but documentation will also need to be updated.

Documentation is for a sure a thing that has to be done. I will try to make commented examples for the integration of ES6Class components into ReactComponentB components and the other way around. From there on we can slowly developing a reasonable documentation.

japgolly:
> I'm not super familiar with the way testing should be done here.

Check out [FunctionalComponentTest.scala](https://github.com/japgolly/scalajs-react/blob/master/test/src/test/scala/japgolly/scalajs/react/FunctionalComponentTest.scala). It's a good example of what we'll need to do here. Most of the tests for this will be the same, i.e. just creating a component, rendering it and comparing to expected html. (There are also other tests in that directory if you're interested.)

> JSProps is just the wrapper similar to your WrapObj

Ah k. I haven't read through it all in detail yet. I'm waiting... Actually, I'll continue this at the bottom.

> but the part about the Reuseability iam not sure about yet.

`Reusability` should be a piece of cake. It's only point of interaction with React components is to install a `shouldComponentUpdate` callback.

> What i mean by that is, that we should ensure that the `TopNode` is not restricted to the *DOM* in any point. That's the only way we can ensure that we are able to integrate react-naive in the future.

That's a very good point!

> Documentation is for a sure a thing that has to be done. I will try to make commented examples for the integration of ES6Class components into ReactComponentB components and the other way around. From there on we can slowly developing a reasonable documentation.

Interesting. That's not a perspective I would personally care about but I can see how that would appeal to people already used to React without Scala. Cool.


Ok, so here's what I think we should do: let's split this entire feature up into small chunks and focus on one chunk at a time. We can add a new branch and merge pieces one at a time. First let's upgrade to Scala 0.6.5 and nothing else (because it's a noisy upgrade with `js.native` everywhere). Second let's add the bare minimum for this component and unit tests that it rendered the expected result given certain Scala-based props & state. I've setup a `topic/es6` branch. Sound good?

xsistens:
Hey David,
i continued my work on this feature a bit. I will give you a short overview of what i was able to accomplish:

- Removed ```JSProps``` and replaced it with your ```WrapObj```
- Reused a lot of your code. ES6 Components now produce a ```ReactComponentC.ReqProps``` and from there on you can handle them the same way as the other Components.
- Implemented similar behaviour for ```Ref```s
- Implemented similiar ```Callback``` behavior for ```setState```
- Wrote examples, where ES6 components and ReactComponentB components interact with each other.

So i think a lot of the things you defined as the first chunk are already working. All I/we need to do is write the missing tests for it. 

I like your suggestion about the further procedure. I can write a PR for your master branch with only the upgrade to scala.js 0.6.5. And then i can write a PR with the current status of my es6 branch for the new topic. This offers way better integration into the already existing code. From there on we can implement the missing tests step by step. Maybe we even get help from other dedicated developers :) After the test coverage for this is acceptable we can start thinking about the integration of the ```Reusability``` part. Does this sounds like a plan to you?

hamero:
+1 for this change. I'm already using this to pass scalajs defined react component to pure js react component.

japgolly:
Hey Christian,

What you're doing sounds fantastic. It makes a lot of sense that to arrive at a `ReactComponentC.xxx` class because they are just wrappers for Scala-based components regardless of how the components are created.

Your suggestion sounds good to me. Let's do it. :+1:

PR #204, opened 2015-10-16
I guess a lot of people will copy this section, and i needed to speci…

elacin:
…fy `dependsOn` in my build for it to work


japgolly:
Ah, that's helpful. Thanks!

Issue #198, opened 2015-10-12
CSSTransitionGroup timeout props


zv3:
Was checking the React docs (https://facebook.github.io/react/docs/animation.html) on CSSTransitionGroup and wanted to play with the transitionEnterTimeout and transitionLeaveTimeout because somehow removed elements (from a collection in the state) are not removed from the DOM after re-rendering. This is on v0.10 and React 0.14

Thought that playing with the timeout values would fix this but seems we can only set transitionAppear/transitionLeave boolean props

japgolly:
Ah right, sounds like we just need to add `transitionEnterTimeout` and `transitionLeaveTimeout` params and we should be sweet.

zv3:
Excellent, I can open a PR if you like, I think I can do that :smiley:

japgolly:
Sure! That would be helpful :smile:

zv3:
Submitted PR #205 and tested the added props, all working :+1:

japgolly:
Closed via #205

PR #205, opened 2015-10-16
Addons: Added support for {Appear,Enter,Leave}Timeout optional props to ReactCSSTransitionGroup case class

zv3:
PR for #198 

https://facebook.github.io/react/docs/animation.html

https://github.com/facebook/react/blob/master/src/addons/transitions/ReactCSSTransitionGroup.js#L53-L58


japgolly:
LGTM, cheers mate!

PR #212, opened 2015-10-26
Basic ES6 Component without Tests and Reuseability Part

xsistens:
Here is my first PR for the es6 topic branch. Everything written here is working so far. I also added some basic tests to the ```CoreTest```s file, but even the unmodified tests gave me some errors. Probably because of my phantomjs setup or something like this, had not enough time to investigate into this. In the ghpages project are some examples on how es6 components can be used with ```ReactComponentB``` components. 

One thing that bothers me a bit and i'm not really sure how to handle it is the type parameter ```N <: TopNode``` for the reason of the DOM dependency. I think if we can get rid of this in some way, this branch could also be used from @chandu0101 as a base for his sri project. I think there has to be a solution for it, at the latest when you start separating the projects into the core and dom projects like facebook was doing it. Maybe you can pull out something in this regard :)

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/Addons.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/Addons.scala
@@ -67,13 +67,16 @@ object Addons {
    * https://facebook.github.io/react/docs/perf.html
    */
   @JSName("React.addons.Perf")
+  @js.native
   object Perf extends js.Object {
     type Measurements = js.Array[Measurement]
 
+    @js.native
     sealed trait Measurement extends js.Object {
       val totalTime: Double = js.native
     }
 
+    @js.native
     sealed trait Report extends js.Object
 
     def start(): Unit = js.native
--- a/core/src/main/scala/japgolly/scalajs/react/CallbackOption.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/CallbackOption.scala
@@ -9,7 +9,7 @@ object CallbackOption {
     new CallbackOption(cb.toScalaFn)
 
   def empty: CallbackOption[Unit] =
-    pure(())
+    CallbackOption(CallbackTo pure someUnit)
 
   def pure[A](a: A): CallbackOption[A] =
     CallbackOption(CallbackTo pure Some(a))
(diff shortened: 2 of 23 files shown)


japgolly:
I thought we were going to submit just the Scala.JS v0.6.5 upgrade first to master as a single commit. Maybe there was a misunderstanding - no matter. Could we do that first, please? Then I'll tweak a little bit, I'll merge master into es6 and then we can rebase this so that the PR is clearer and easier to review.

japgolly:
Regarding the `TopNode` and splitting stuff, hmm, let me stew on it a little. I haven't looked at the native stuff at all. Does it have `findDOMNode`/`getDOMNode` too? If not, does it have anything similar?

xsistens:
I PR the scalajs version change in a minute to master. Should i resend a PR for the topic/es6 branch without the master merge or how should i do it now?

japgolly:
After we merge just the Scala.JS upgrade, I'll tweak a little bit, I'll merge master into es6 and then we can rebase this so that the PR is clearer and easier to review.

So just pause this PR temporarily and I'll ask you to rebase when ready.

xsistens:
> Regarding the TopNode and splitting stuff, hmm, let me stew on it a little. I haven't looked at the native stuff at all. Does it have findDOMNode/getDOMNode too? If not, does it have anything similar?

As far as i know, there is no similar thing in react-native. I also asked @chandu0101, because I think he has the most experience with react-native. It only exists in the *ReactDOM* part of react.

japgolly:
*groan*

xsistens:
It's deprecated and will be removed everywhere else in the next release too.

japgolly:
That's only `getDOMNode` but not `findDOMNode`. They added `find` as a
replacement for `get`. They just moved it.

xsistens:
Yeah, ```findDOMNode``` replaces ```getDOMNode``` in the ```ReactDOM``` module. But there will be no more ```findDOMNode``` or ```getDOMNode``` in the component level API. Thats what I meant with it.

xsistens:
Sorry I did a big mistake by accidentally using the wrong branch! I will open a new one with the correct Files.

japgolly:
Oh? Ok, cool. `topic/es6` is ready for you now :)

PR #213, opened 2015-10-26
Scala.js version update to 0.6.5

xsistens:


japgolly:
Gee that's a lot of `@js.native` everywhere! I'll take care of this tomorrow morning and get back to you.

xsistens:
yeah, you have to. Everywhere ```js.native``` is used we have to put it. More precisely everywhere javascript access is involved it has to be annotated with ether ```@js.native``` or ```@ScalaJSDefined```.
Ok, have a nice day! Or probably night if you are in Australia at the moment :D Cu tomorrow

japgolly:
Yeah I was dreading it when I heard the news for Scala.JS. It must've taken you some time. You are correct indeed! It's night time here and the aftermath of dinner is rotting in the kitchen sink. Better get away from the computer and clean up. Talk tomorrow! :D

PR #217, opened 2015-10-28
ES6 component support

xsistens:
This PR should finally do it. I'm sorry, I'm not super experienced with git and github.

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/Addons.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/Addons.scala
@@ -68,6 +68,7 @@ object Addons {
    */
   @js.native
   @JSName("React.addons.Perf")
+  @js.native
   object Perf extends js.Object {
     type Measurements = js.Array[Measurement]
 
--- a/core/src/main/scala/japgolly/scalajs/react/React.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/React.scala
@@ -33,7 +33,8 @@ trait React extends Object {
    * Create and return a new `ReactElement` of the given type. The type argument can be either an html tag name string
    * (eg. 'div', 'span', etc), or a [[ReactClass]] (created via [[React.createClass]]).
    */
-  def createElement[P,S,B,N <: TopNode](t: ReactClass[P,S,B,N]): ReactComponentCU[P,S,B,N] = js.native
+  def createElement[P,S,B,N <: TopNode](t: ReactClass[P,S,B,N], props: WrapObj[P], children: ReactNode*): ReactElement = js.native
+
   /**
    * Create and return a new `ReactElement` of the given type. The type argument can be either an html tag name string
    * (eg. 'div', 'span', etc), or a [[ReactClass]] (created via [[React.createClass]]).
--- a/core/src/main/scala/japgolly/scalajs/react/ReactComponent.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/ReactComponent.scala
@@ -1,0 +1,141 @@
+package japgolly.scalajs.react
+
+import japgolly.scalajs.react
+
+import scala.scalajs.js
+import scala.scalajs.js.UndefOr
+import scala.scalajs.js.annotation.{JSName, ScalaJSDefined}
+
+@js.native
+@JSName("React.Component")
+private[react] class ReactJSComponent[P, S] extends js.Object {
+  val displayName: String = js.native
+
+  var refs: RefsObject = js.native
+
+  var context: js.Dynamic = js.native
+  @JSName("state")
+  private[react] var _state: WrapObj[S] = js.native
+
+  def componentDidMount(): Unit = js.native
+
+  def componentWillUnmount(): Unit = js.native
+
+  @JSName("forceUpdate")
+  private[react] def _forceUpdate(callback: js.UndefOr[js.Function]): Unit = js.native
+
+  @JSName("componentWillReceiveProps")
+  private[react] def _componentWillReceiveProps(nextProps: WrapObj[P]): Unit = js.native
+
+  @JSName("shouldComponentUpdate")
+  private[react] def _shouldComponentUpdate(nextProps: WrapObj[P], nextState: WrapObj[S]): Boolean = js.native
+
+  @JSName("componentWillUpdate")
+  private[react] def _componentWillUpdate(nextProps: WrapObj[P], nextState: WrapObj[S]): Unit = js.native
+
+  @JSName("componentDidUpdate")
+  private[react] def _componentDidUpdate(prevProps: WrapObj[P], prevState: WrapObj[S]): Unit = js.native
+
+  @JSName("props")
+  private[react] def _props: WrapObj[P] with PropsMixedIn = js.native
+
+  @JSName("setState")
+  private[react] def _setState(s: WrapObj[S], callback: UndefOr[js.Function]): Unit = js.native
+
+  @JSName("setState")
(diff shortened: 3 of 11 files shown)


core/src/main/scala/japgolly/scalajs/react/React.scala
```diff
@@ -33,7 +33,8 @@ trait React extends Object {
    * Create and return a new `ReactElement` of the given type. The type argument can be either an html tag name string
    * (eg. 'div', 'span', etc), or a [[ReactClass]] (created via [[React.createClass]]).
    */
-  def createElement[P,S,B,N <: TopNode](t: ReactClass[P,S,B,N]): ReactComponentCU[P,S,B,N] = js.native
+  def createElement[P,S,B,N <: TopNode](t: ReactClass[P,S,B,N], props: WrapObj[P], children: ReactNode*): ReactElement = js.native
```
japgolly: We can't have this.
* The types are lose in the output.
* `WrapObj` is an internal mechnism and not part of React's facade nor something that should be exposed to users.

core/src/main/scala/japgolly/scalajs/react/Addons.scala
```diff
@@ -68,6 +68,7 @@ object Addons {
    */
   @js.native
   @JSName("React.addons.Perf")
+  @js.native
```
japgolly: Unnecessary

japgolly:
Sorry, this PR makes me uneasy. I'm gonna have to come back to this later to think and be more thorough than I have time to right now.

> I'm not super experienced with git and github.
Hehe you will be soon with all of this experience :grinning:

core/src/main/scala/japgolly/scalajs/react/ReactComponent.scala
```diff
@@ -0,0 +1,141 @@
...
+  @JSName("_forceUpdate")
+  def forceUpdate(callback: Callback = Callback.empty): Unit =
+    _forceUpdate(callback.toJsCallback)
+
+  @JSName("componentWillUpdate")
+  override def _componentWillUpdate(nextProps: WrapObj[P], nextState: WrapObj[S]): Unit =
+    componentWillUpdate(nextProps.v, nextState.v)
+
+  @JSName("_componentWillUpdate")
+  def componentWillUpdate(nextProps: => P, nextState: => S): Unit = ()
```
japgolly: Non-strict args? Would this really work? I imagine there'd be some kind of interop problem, no?

core/src/main/scala/japgolly/scalajs/react/Addons.scala
xsistens: Sorry, was a mistake from all the rebasing. I will remove that.

core/src/main/scala/japgolly/scalajs/react/React.scala
xsistens: I can remove that. Its was only needed for my experimenting with scalajs-relay. It is not needed for the ES5 classes in scalajs-react at all. I will revert it to the old code.

japgolly:
@xsistens I was hoping to go over this this weekend but wasn't able to. I just want you to know you are not forgotten, I've just got a few things going irl.

xsistens:
> @xsistens I was hoping to go over this this weekend but wasn't able to. I just want you to know you are not forgotten, I've just got a few things going irl.

No problem mate, RL always has priority :)! Thanks for the update though. 

At the moment I'm thinking about a way to get rid of the ```WrapObj[P]``` to access the props.
I think its one of the main problems, why utilizing other complex react libraries is kinda impossible at the moment. As soon as you loose control over the instantiating process of the react components (the process where you pass the props to the components) there is at the moment no way to access the props with type safety. I guess a more flexible way to define how the the props are getting de/en-coded would be really beneficial. I was thinking about two functions like 
```
def encode(props: P): js.Dynamic
def decode(js.Dynamic): P
```
that you can pass over. The default Implementation for this could be the same as it is right now, that you put the whole props in a attribute *"v"*. But if you need to control it you can write this two functions yourself. 

To give you an example:
If you want to write a facade for relay, don't want to dig too deep into the relay code and just provide a facade for the basic access points. That's why you don't have the control how the react components you define yourself are getting created with the props relay is gathering for you. So relay would always override your props without wrapping them under a *"v"* attribute and if you want to access them via *"v"* they will be missing.

Maybe we can find a solution for that, so that the community can utilize other react libraries like relay or react-router, which is impossible with the current implementation.

japgolly:
Hi @xsistens, I'll try to get this back on track. I originally wanted to work on this in tandem with you but I simply don't have the luxury given a number of life-circumstances of mine. So, if it's ok with you, I'll just guide the direction until this is ready for merging.

To get started, where are you at with this now? Both in terms of code and motivation?

If you're willing, I'll start creating a list of boxes that will need to be ticked to get this merged. Things like:

* Must use the `Callback` type for callbacks.
* Access to `this` must be type-safe and enforce a number of rules scattered throughout the React documentation (that aren't always clear). This means preventing access to certain methods and changing type signatures depending on where `this` is accessed. (Note: This is currently achieved in the library already - that can be a reference.) This will be a challenge.
* Must play well with all the tools in `.extra`.
* Must be unit tested.
* Must be documented (in `doc/`).
* After creation, must be indistinguishable from components created via the builder (i.e. must fit the same interfaces/types).

xsistens:
Hey @japgolly, I'll try to give you a short feedback on the status of the project i achieved in my fork. 

> - Must use the Callback type for callbacks.

I already tried to mirror the rest of scalajs-react and use Callbacks whereever they are used in the existing project. So thats not a big deal.

> - Access to this must be type-safe and enforce a number of rules scattered throughout the React documentation (that aren't always clear). This means preventing access to certain methods and changing type signatures depending on where this is accessed. (Note: This is currently achieved in the library already - that can be a reference.) This will be a challenge.

This is pretty much untouched and like you said takes some time (and as you know time is a thing we all need a bit more of :) ). What I want to say with it is, that right now I dont have the time to investigate deeper for that goal to be achieved, but when enought free time is there I 'am willed to find a solotion for that.

> - Must play well with all the tools in .extra.

I adopted some changes to the .extra project, but I'am  not super familiar with all the stuff in there, thats why I have not tested what is working and what is not. Maybe after writing some docs, how to use ES6 components someone else who has a better understanding of all the things in the extra project can investigate deeper into this. One thing I saw that is used for some extra stuff is the ```.configure``` method which is not present for ES6 right now. I have no idea how to accomplish that for ES6 at the moment.

> - Must be unit tested.

I already wrote some tests, but I have the problem that the tests are not executing correctly (probably a problem with my setup). I have to find a solution for that and then I can provide some tests. But with my spear time I would be greatful for every help with it that comes from the community.

> - Must be documented (in doc/).

I can provide a first introduction there in how to use the ES6 components (with the code right now) for the purpose, that other people could check it out, get started and maybe help with some problems like test or so. Especially when you want all the bullet-points (and with some future changes to scala-js) this docs will have to change.

> - After creation, must be indistinguishable from components created via the builder (i.e. must fit the same interfaces/types).

I tried to achive this as much as possible, but there are some serious problems with that. For example ES6 components have no backend class, because it is not needed. So when accessing components via a ref you can access their methods, props and state without calling the ```.backend``` function first. What I mean is, that there will be some small differences between the components (like there are some differences with JSComponents as well). But after creating a ES6 component you will also get a ```ReactComponentC``` and from there on you can treat them equally.


So to sum it up, some of the needed code is already there, the motivation is high but same like you, the time is a bit lacking. I would like to see a way where we could present something like a beta branch to the rest of the community, where we can achieve the bullet-points you mentioned step by step. 

I hope my english is not to horrible and I could answer some of your questions.

PR #231, opened 2015-11-26
Added a boolean extension ?= for ReactNode

ochrons:


ochrons:
Oh well, looks like it doesn't work that well when implemented this way...

PR #250, opened 2016-01-19
Adjust dom.alert references to dom.window.alert

easel:
- dom.alert will disappear in scala-js-dom v0.9.0


japgolly:
I'm interested in knowing more about this. Why is it so? Which issue in scala-js-dom is this? I was having a look around and didn't see anything that would remove `dom.alert`.

easel:
Not sure, TBH, but I built against the snapshot and it wouldn't compile. I didn't chase down where it got removed.

japgolly:
>I built against the snapshot and it wouldn't compile

Oh I get you now. Right, right, well let's hold off on this change until 0.9.0 final is release. Then we can update the dep and fix any breakages all at once.

PR #251, opened 2016-01-19
Provide safer BaseUrl.fromWindowLocation methods

### LINKED ISSUE #249: router blows up with UBE on "some" ie versions
Root cause is referencing window.location.origin, which isn't well supported on IE.

I'm working on getting this fixed upstream, so this should get fixed automatically with an update to scala.js or scala-js.dom, but for now you can also work around it with the polyfill from  https://github.com/scala-js/scala-js-dom/issues/201

easel:
- IE doesn't properly support BaseUrl.fromWindowOrigin, so until the scala-js-dom facade
  represents the possible undefined state for dom.window.location.origin, one can use this
  method which can be used a fallback once the dom type is updated

Fixes #249

### DIFF
--- a/doc/changelog/0.10.5.md
+++ b/doc/changelog/0.10.5.md
@@ -10,3 +10,4 @@
 * Add new Router DSL:
   * `remainingPath` - Captures the (non-empty) remaining portion of the URL path.
   * `remainingPathOrBlank` - Captures the (potentially-empty) remaining portion of the URL path.
+* Replace BaseUrl.fromWindowOrigin with a more IE-friendly implementation that does not use location.origin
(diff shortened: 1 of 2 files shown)


japgolly:
A few small things:

1. Please update the changelog in `doc/changelog/`

2. This should replace the `fromWindowOrigin` implementation cos that's what this is doing. No need for new methods.

3. Sorry I'm going to be pedantic here. I prefer this implementation:

   ```scala
    val l = dom.window.location
    var url = l.protocol + "/" + l.hostname
    if (!l.port.matches("^(?:80)?$"))
      url += ":" + l.port
    BaseUrl(url)
    ```
    
  [Apparently](http://stackoverflow.com/questions/13167302/did-ie10-change-the-definition-of-window-location-port) some versions of IE return `"80"` for port. Secondly, I think there's a hidden cost to using string interpolation and in this case I found it wasn't helping readability - quite the opposite IMO.

Cheers!

easel:
Ok, I think the latest covers everything. Thanks!

japgolly:
Cool, thanks! Merged.

PR #253, opened 2016-01-22
Create proper urls in fromWindowOrigin

easel:


easel:
Looks like we missed a crucial character in this fix. You get some real nice redirect loops without the second "/" =p

japgolly:
Hahaha, we suck! Thanks.

easel:
Yeah, I was pretty impressed with myself for missing that. Thanks for merging!

japgolly:
Even more thanks for noticing and fixing :smile:

japgolly:
Wait shouldn't it be "://" ?

easel:
Nope. Just //. Don't ask me why. I tried it a bunch of ways but the // is what I had in the string interpolation which I copied from the js polyfill and it works. ://, empty and / all create redirect loops.

japgolly:
Wow, weird. Thanks for looking at a polyfill. I'd really love a unit test but don't want to mock out all the `window` fields. Oh well; it should be one of those things we never need to touch again once written.

japgolly:
Btw, you may already know this but adding `.logToConsole` to your router config makes understanding exactly what it does and why much clearer. FYI.

easel:
Yeah, I've definitely gotten some good mileage out that!

easel:
Incidentally if you want to see some good head-slappers, check out the upstream ticket in scala-js-dom for the issue that caused this. Microsoft has open tickets about the missing .origin property, apparently it works on IE11 Win7, IE11 patched Win10, but not on older IE11 on Win10. Crazy.

Also, the polyfills I found all just mutate the property onto the location object. For some reason that worries me that it won't be kept up to date properly with html5 pushstate, so I like the implementation we have that is a function better.

PR #257, opened 2016-02-15
Example using the WebSockets API

cquiroz:
This PR contains a Scala ReactJS example using WebSockets connecting to the well-known WebSockets echo service at http://www.websocket.org/echo.html

I'm still learning my way around scalajs-react. I welcome any comments on my code as it may be not as idiomatic as it should be, in particular I'm not sure about having the `WebSocket` object inside the `State` class.

Does it need some type of copyright note?

### DIFF
--- a/gh-pages/src/main/scala/ghpages/examples/WebSocketsExample.scala
+++ b/gh-pages/src/main/scala/ghpages/examples/WebSocketsExample.scala
@@ -1,0 +1,99 @@
+package ghpages.examples
+
+import ghpages.GhPagesMacros
+import ghpages.examples.util.SingleSide
+import japgolly.scalajs.react._, vdom.prefix_<^._
+
+import org.scalajs.dom.{WebSocket, MessageEvent, Event, CloseEvent, ErrorEvent}
+
+object WebSocketsExample {
+
+  def content = SingleSide.Content(source, main())
+
+  lazy val main = addIntro(WebSocketsApp, _(
+    s"Echo messages with WebSockets using ReactJS"))
+
+  val source = GhPagesMacros.exampleSource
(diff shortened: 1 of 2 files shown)


japgolly:
Very cool; thanks @cquiroz! There are a few small improvements I can make, just minor things. I'll merge this tomorrow and then make the adjustments if that's cool with you.

> Does it need some type of copyright note?

I don't believe so.

cquiroz:
Feel free to update it as needed. I'm looking forward to see the improvements
BTW, very cool system for the examples, I guess they are compiled as part of the travis check?

japgolly:
Hi @cquiroz. I made a plethora of small changes in a12e0bccff90dcc1cdb98f16e1d5f7c1122a0f46; I hope you don't mind. I put some comments in there to explain the changes. In some cases there is an inline comment to explain the change.

> BTW, very cool system for the examples, I guess they are compiled as part of the travis check?

Yeah they're compiled as part of Travis to make sure they compile but they aren't published. Publication is manual and on the `gh-pages` branch (if that's what you meant).

Finally, thanks again for the PR. Always nice to see more examples added :) :beer:

cquiroz:
That's very cool, Somewhat the comment about form is interesting, I'd think that for usability reasons having it as a form should be the preferred way, though I was hit by the bug of not preventing the event
Nice to see the proper way to link the WS callbacks to react

japgolly:
Hey @cquiroz, I just made another change because the example broke online. https trying to connect to ws failed and we forgot to catch it. So I've shown how to catch errors now and changed to wss. 525f34e

cquiroz:
Looks very good

PR #283, opened 2016-05-09
`assert` had to be `assertHtml` in Test Documentation

i-am-the-slime:
This seems to fix the example. I am having trouble with my component that doesn't take any props for that example so any help with that would be nice.

### DIFF
--- a/doc/TESTING.md
+++ b/doc/TESTING.md
@@ -133,10 +133,10 @@ ComponentTester(Example)("First props") { tester =>
   assertHtml("First props", 0)
 
   setState(2)
-  assert("First props", 2)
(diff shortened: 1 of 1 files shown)


japgolly:
Thanks. SIlly typos!

Are you still having problems testing your component? Depending on what you want to test with your component, it's usually easy. Make a look through this project's unit tests to see a mish-mash of different testing approaches I've used over the years.

If what you want to test is non-trivial, I suggest using [TestState](https://github.com/japgolly/test-state). It's how I test the majority of my real project's components. A full demo using it with scalajs-react is [here](https://github.com/japgolly/test-state/tree/master/example-react).

PR #285, opened 2016-05-25
Fix link in TESTING.md

OlivierBlanvillain:


doc/TESTING.md
```diff
@@ -218,7 +218,7 @@ object ExampleTest extends TestSuite {
 
 `DebugJs`
 =========
-[DebugJs](src/main/scala/japgolly/scalajs/react/test/DebugJs.scala) is a dumping ground for functionality useful when testing raw JS.
+[DebugJs](https://github.com/japgolly/scalajs-react/blob/master/test/src/main/scala/japgolly/scalajs/react/test/DebugJs.scala) is a dumping ground for functionality useful when testing raw JS.
```
japgolly: Ah thank you. We should use relative links though so it looks like it's missing the `../test/` bit.

OlivierBlanvillain: Was not sure which one would be more robust :smile:

japgolly:
Beauty! Thanks @OlivierBlanvillain

PR #307, opened 2016-10-06
Advice on how to change the state from outside

MikaelMayer:


japgolly:
Good call. Thank you.
I might also follow up with an explicit link to the *Callbacks* section for more detail.

Issue #299, opened 2016-07-30
Support onClickCapture and friends


jtcwang:
From the react docs:

> The event handlers below are triggered by an event in the bubbling phase. To register an event handler for the capture phase, append Capture to the event name; for example, instead of using onClick, you would use **onClickCapture** to handle the click event in the capture phase.

Should we add support for this as well? I think we can just add the relevant code in `HtmlAttrs.scala` - Do you see any problems with this?

japgolly:
What exactly is the distinction? Is capture/bubbling like saying will-fire/has-fired?

What is your use case for which you need this? (Not for PR purposes, I'd just like to learn and understand.)

Adding that to `HtmlAttrs` sounds fine.

jtcwang:
Here's a good explanation: http://stackoverflow.com/questions/4616694/what-is-event-bubbling-and-capturing

So `onClickCapture` that's bound on a container element will allows you to handle events in the container before  the inner elements receives the event. This is useful in many cases when you want to stop propagation of the event to inner elements if some logic determines that you don't want the inner element to see the event at all.

jtcwang:
So I looked at the React source code and here is the full list, as salvaged from

japgolly:
Sorry I left this hanging for the last 2 weeks. Life be very busy these days.

Yes, I'd be happy for you to add those events. You have two options (or you can do both) - I don't mind.

1) Add it to `master` and I'll do a release that will be the last non-bugfix release for v0.x.y

2) Add it to `topic/neo` branch in [this file](https://github.com/japgolly/scalajs-react/blob/topic/neo/neo/src/main/scala/japgolly/scalajs/react/vdom/HtmlAttrs.scala). If you do this, all the fields are in alphabetical order so please follow suit.

jtcwang:
I'm not in dire need of this. I'll add it to `topic/neo` branch

jtcwang:
Since this is merged into `topic/neo`. I think we can close this.

japgolly:
No worries. If you didn't see, I dropped a msg in gitter to say that unfortunately I've had to postpoone work on neo. I'll pick it again in January.

jtcwang:
Yeap I saw that. Do appreciate the efforts you put into this project regardless!

PR #308, opened 2016-10-09
Added *Capture variant of react events (e.g. onClickCapture)

jtcwang:
For #299. As collected from React's source code

There are no *Capture variants according to React's [official doc](https://facebook.github.io/react/docs/events.html).
```
The onMouseEnter and onMouseLeave events propagate from the element 
being left to the one being entered instead of ordinary bubbling and do not have 
a capture phase.
```

### DIFF
--- a/neo/src/main/scala/japgolly/scalajs/react/vdom/HtmlAttrs.scala
+++ b/neo/src/main/scala/japgolly/scalajs/react/vdom/HtmlAttrs.scala
@@ -545,11 +545,19 @@ trait HtmlAttrs {
 
   final lazy val onAbort = Attr.Event.base("onAbort")
 
+  final lazy val onAbortCapture = Attr.Event.base("onAbortCapture")
+
   final lazy val onAnimationEnd = Attr.Event.animation("onAnimationEnd")
+
+  final lazy val onAnimationEndCapture = Attr.Event.animation("onAnimationEndCapture")
 
   final lazy val onAnimationIteration = Attr.Event.animation("onAnimationIteration")
 
+  final lazy val onAnimationIterationCapture = Attr.Event.animation("onAnimationIterationCapture")
+
   final lazy val onAnimationStart = Attr.Event.animation("onAnimationStart")
+
+  final lazy val onAnimationStartCapture = Attr.Event.animation("onAnimationStartCapture")
 
   final lazy val onBeforeInput = Attr.Event.base("onBeforeInput")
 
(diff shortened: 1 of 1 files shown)


neo/src/main/scala/japgolly/scalajs/react/vdom/HtmlAttrs.scala
```diff
@@ -545,21 +545,33 @@ trait HtmlAttrs {
...
+
   final lazy val onAnimationIteration = Attr.Event.animation("onAnimationIteration")
 
+  final lazy val onAnimationIterationCapture = Attr.Event.animation("onAnimationIterationCapture")
+
   final lazy val onAnimationStart = Attr.Event.animation("onAnimationStart")
 
+  final lazy val onAnimationStartCapture = Attr.Event.animation("onAnimationStartCapture")
+
   final lazy val onBeforeInput = Attr.Event.base("onBeforeInput")
```
japgolly: No `onBeforeInputCapture`?

neo/src/main/scala/japgolly/scalajs/react/vdom/HtmlAttrs.scala
```diff
@@ -574,6 +586,8 @@ trait HtmlAttrs {
     */
   final val onClick = Attr.Event.mouse("onClick")
 
+  final val onClickCapture = Attr.Event.mouse("onClickCapture")
+
   final lazy val onCompositionEnd = Attr.Event.composition("onCompositionEnd")
```
japgolly: No `onCompositionEndCapture`? Is this omission deliberate? Do all events have capturing versions?

japgolly: I just had a thought: It might be a good idea to have the event methods return a tuple of events `(normal, capture)` so that 1) every event always has a capture version, and 2) it's less boilerplate.

Something like:
```scala
final lazy val (onCompositionEnd, onCompositionEndCapture) = Attr.Event.composition("onCompositionEnd")
```

WDYT?

neo/src/main/scala/japgolly/scalajs/react/vdom/HtmlAttrs.scala
jtcwang: I was working off the list I posted in #299 (which is directly scavenged from the React source code). I will check it again and see if there are any new ones or whether I missed any.

neo/src/main/scala/japgolly/scalajs/react/vdom/HtmlAttrs.scala
japgolly: Ah got it. I wasn't sure if all events have capture versions. Ok , no worries. LGTM and thanks!

PR #309, opened 2016-10-11
Support for namespaces in JsComponent name.

jeroentervoorde:
Small change for neo. Supports using a js React component in a different namespace than global.
Also throws a more descriptive error message when the component constructor does not exist.

For example: 
```
JsComponent.byName[Props, ChildrenArg.Varargs, Props]("ReactLeaflet.Map")
```


japgolly:
Great idea, that's really cool! Could you please add a test to cover this?

jeroentervoorde:
I added a test and a check to verify the resolved object is actually a js.Function,

japgolly:
Very nice, thank you!

PR #310, opened 2016-10-19
Upgrade to Scala.js 0.6.13

julienrf:
I wasn’t able to run the tests because I don’t have phantomjs but everything compiles fine without any warning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,5 +50,5 @@ Includes a router, testing utils, performance utils, more.
 ##### Requirements:
 * React 15+
 * Scala 2.11.n
-* Scala.JS 0.6.8+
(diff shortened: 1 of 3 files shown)


japgolly:
Thanks @julienrf . Because the IR in SJS 0.6.13 has changed, could you please update the minimum SJS version in the `README.md`, and add an entry to the latest changelog.

julienrf:
@japgolly Done! I was not sure whether we should cut a 0.11.3 or a 0.12.3, since the API did not change but the requirements changed…

japgolly:
Yeah, I think 0.11.3 is appropriate here. Ok thanks @julienrf! Let's merge it.

PR #320, opened 2016-12-02
Migrate to scalajs-bundler

julienrf:
I wanted to play with scalajs-bundler in the context of a large Scala.js facade, so I naturally went to scalajs-react, which is one of the most used facades.

It turns out that it works well with scalajs-bundler. In my opinion the builds of projects that use scalajs-react become simpler with this PR. I will highlight some parts of the PR by commenting in the diff.

A drawback of scalajs-bundler is that a concrete facade can not be compatible with both `npmDependencies` and `jsDependencies`, so existing users will be forced to move to scalajs-bundler. A solution to this problem could be to isolate the facade interfaces into separate, small sbt sub-projects, and then have two different concrete facades (one that would work with `npmDependencies` and the other with `jsDependencies`).

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/ReactDOM.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/ReactDOM.scala
@@ -1,9 +1,12 @@
 package japgolly.scalajs.react
 
 import org.scalajs.dom
-import scala.scalajs.js
-import js.{ThisFunction, ThisFunction0, Object}
 
+import scala.scalajs.js
+import js.{Object, ThisFunction, ThisFunction0}
+import scala.scalajs.js.annotation.JSImport
+
+@JSImport("react-dom", JSImport.Namespace)
 @js.native
 object ReactDOM extends ReactDOM
 
@@ -118,6 +121,7 @@ trait ReactDOM extends Object {
   def findDOMNode[N <: TopNode](component: CompScope.Mounted[N]): N = js.native
 }
 
+@JSImport("react-dom/server", JSImport.Namespace)
 @js.native
 object ReactDOMServer extends ReactDOMServer
 
--- a/doc/USAGE.md
+++ b/doc/USAGE.md
@@ -20,33 +20,13 @@ It is expected that you know how React itself works.
 Setup
 =====
 
-1. Add [Scala.js](http://www.scala-js.org) to your project.
+1. Add [scalajs-bundler](https://scalacenter.github.io/scalajs-bundler/getting-started.html) to your project.
 
 2. Add *scalajs-react* to SBT:
 
   ```scala
   // core = essentials only. No bells or whistles.
   libraryDependencies += "com.github.japgolly.scalajs-react" %%% "core" % "0.11.3"
-
-  // React JS itself (Note the filenames, adjust as needed, eg. to remove addons.)
-  jsDependencies ++= Seq(
-
-    "org.webjars.bower" % "react" % "15.3.2"
-      /        "react-with-addons.js"
-      minified "react-with-addons.min.js"
-      commonJSName "React",
-
-    "org.webjars.bower" % "react" % "15.3.2"
-      /         "react-dom.js"
-      minified  "react-dom.min.js"
-      dependsOn "react-with-addons.js"
-      commonJSName "ReactDOM",
-
-    "org.webjars.bower" % "react" % "15.3.2"
-      /         "react-dom-server.js"
-      minified  "react-dom-server.min.js"
-      dependsOn "react-dom.js"
-      commonJSName "ReactDOMServer")
   ```
 
 Creating Virtual-DOM
--- a/doc/USAGE.md
+++ b/doc/USAGE.md
@@ -29,6 +29,17 @@ Setup
   libraryDependencies += "com.github.japgolly.scalajs-react" %%% "core" % "0.11.3"
   ```
 
+3. Add dependencies to React.js (adjust as needed, eg. to remove addons):
+
+  ~~~ scala
+  npmDependencies in Compile ++= Seq(
+    "react" -> "15.3.2",
+    "react-dom" -> "15.3.2",
+    "react-addons-perf" -> "15.3.2", // Optional
+    "react-addons-css-transition-group" -> "15.3.3" // Optional
+  )
+  ~~~
+
 Creating Virtual-DOM
 ====================
 
--- a/gh-pages/dev.html
+++ b/gh-pages/dev.html
@@ -6,32 +6,11 @@
     <link rel="stylesheet" href="res/bootstrap.css">
     <link rel="stylesheet" href="res/app.css">
     <link rel="stylesheet" href="res/default.css">
-    <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/react/15.0.1/react-with-addons.js"></script>
-    <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/react/15.0.1/react-dom.js"></script>
     <script type="text/javascript" src="res/highlight.pack.js"></script>
 </head>
 <body>
 
-<script>
-    function jsonp(url, callback) {
-        var callbackName = 'jsonp_callback_' + Math.round(100000 * Math.random());
-        window[callbackName] = function(data) {
-            delete window[callbackName];
-//            document.body.removeChild(script);
-            callback(data);
-        };
+<script type="text/javascript" src="target/scala-2.11/gh-pages-fastopt-bundle.js"></script>
 
-        var script = document.createElement('script');
-        script.src = url + (url.indexOf('?') >= 0 ? '&' : '?') + 'callback=' + callbackName;
-        document.body.appendChild(script);
-    }
-</script>
-
-<script type="text/javascript" src="target/scala-2.11/gh-pages-fastopt.js"></script>
-
-<script type="text/javascript">
-  ((typeof global === "object" && global &&
-         global["Object"] === Object) ? global : this)["ghpages"]["GhPages"]().main();
-</script>
 </body>
 </html>
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -2,8 +2,9 @@ import sbt._
 import Keys._
 import com.typesafe.sbt.pgp.PgpKeys._
 import org.scalajs.sbtplugin.ScalaJSPlugin
-import ScalaJSPlugin._
 import ScalaJSPlugin.autoImport._
+import scalajsbundler.ScalaJSBundlerPlugin
+import ScalaJSBundlerPlugin.autoImport._
 
 object ScalajsReact {
 
@@ -25,7 +26,7 @@ object ScalajsReact {
   val clearScreenTask = TaskKey[Unit]("clear", "Clears the screen.")
 
   def commonSettings: PE =
-    _.enablePlugins(ScalaJSPlugin)
+    _.enablePlugins(ScalaJSBundlerPlugin)
       .settings(
         organization       := "com.github.japgolly.scalajs-react",
         homepage           := Some(url("https://github.com/japgolly/scalajs-react")),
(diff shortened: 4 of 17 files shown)


doc/USAGE.md
```diff
@@ -20,33 +20,13 @@ It is expected that you know how React itself works.
...
-      /         "react-dom.js"
-      minified  "react-dom.min.js"
-      dependsOn "react-with-addons.js"
-      commonJSName "ReactDOM",
-
-    "org.webjars.bower" % "react" % "15.3.2"
-      /         "react-dom-server.js"
-      minified  "react-dom-server.min.js"
-      dependsOn "react-dom.js"
-      commonJSName "ReactDOMServer")
```
julienrf: There is no need to add these lines to the build anymore. Depending on the facade is enough.

gh-pages/dev.html
```diff
@@ -6,32 +6,11 @@
     <link rel="stylesheet" href="res/bootstrap.css">
     <link rel="stylesheet" href="res/app.css">
     <link rel="stylesheet" href="res/default.css">
-    <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/react/15.0.1/react-with-addons.js"></script>
-    <script type="text/javascript" src="//cdnjs.cloudflare.com/ajax/libs/react/15.0.1/react-dom.js"></script>
```
julienrf: These lines are not needed anymore, the dependencies are bundled into a single .js file (referenced at line 13).

test/src/test/resources/sampleReactComponent.js
```diff
@@ -1,4 +1,6 @@
-var SampleReactComponent = React.createClass({
+var React = require('react');
+
+module.exports = React.createClass({
   getInitialState: function() {
     return {num:0,num2:0};
   },
```
julienrf: I turned this code into a proper JS module that does not leak to the global namespace.

japgolly:
Hey @julienrf. Thanks for the PR! scalajs-bundler looks pretty cool.

That being said I think it's too soon to merge this PR into the mainline of scalajs-react. My understanding is that this change will require all library users to use scalajs-bundler. My concern is that scalajs-bundler is new and possibly not production-ready yet, where as a lot of people use scalajs-react in production. Basically it feels a little soon to mandate scalajs-bundler. I'd be happy to create a separate branch for early adopters to use it immediately but as for master, I think it would be better to give it a little more time.

project/Build.scala
```diff
@@ -158,7 +135,14 @@ object ScalajsReact {
     .settings(
       name := "core",
       libraryDependencies ++= Seq(
-        "org.scala-js" %%% "scalajs-dom" % Ver.ScalaJsDom))
+        "org.scala-js" %%% "scalajs-dom" % Ver.ScalaJsDom),
+      npmDependencies in Compile ++= Seq(
+        "react" -> Ver.ReactJs,
+        "react-dom" -> Ver.ReactJs,
+        "react-addons-perf" -> Ver.ReactJs,
+        "react-addons-css-transition-group" -> Ver.ReactJs
```
japgolly: The `react-addons-*` shouldn't be transitive. They're optional. Does `npmDependencies` have a concept of `provided` scope?

project/Build.scala
```diff
@@ -158,7 +135,14 @@ object ScalajsReact {
     .settings(
       name := "core",
       libraryDependencies ++= Seq(
-        "org.scala-js" %%% "scalajs-dom" % Ver.ScalaJsDom))
+        "org.scala-js" %%% "scalajs-dom" % Ver.ScalaJsDom),
+      npmDependencies in Compile ++= Seq(
+        "react" -> Ver.ReactJs,
+        "react-dom" -> Ver.ReactJs,
```
japgolly: Are end-users able to override these versions in their projects? End-users shouldn't be tied to the version of React used in this build.

test/src/test/scala/japgolly/scalajs/react/extra/router/Router2Test.scala
```diff
@@ -136,7 +136,7 @@ object Router2Test extends TestSuite {
 
   override val tests = TestSuite {
     import MyPage2._
-    implicit val base = BaseUrl("file:///router2Demo/")
+    implicit val base = BaseUrl("http://localhost/")
```
japgolly: Why this change?

julienrf:
Hi!

You’re right, it is definitely a better idea to not merge this version to master. It would be great if a version based on scalajs-bundler could exist somewhere, though, as it would make it easier for me to collect feedback on scalajs-bundler :)

project/Build.scala
julienrf: OK, so what we could do is, like you did before, to not define this dependency and require users of the facade to define it by themself (that trick would work for “provided” dependencies too).

julienrf:
@japgolly I added a commit that makes React.js dependencies “provided”.

test/src/test/scala/japgolly/scalajs/react/extra/router/Router2Test.scala
japgolly: That's cool. I'm not sure either. I quickly tried jsdom with scala.js once recently and it wouldn't work at all. Need to try again....

core/src/main/scala/japgolly/scalajs/react/ReactDOM.scala
```diff
@@ -118,6 +121,7 @@ trait ReactDOM extends Object {
   def findDOMNode[N <: TopNode](component: CompScope.Mounted[N]): N = js.native
 }
 
+@JSImport("react-dom/server", JSImport.Namespace)
```
japgolly: This is very cool. It was previously impossible.

project/Build.scala
```diff
@@ -175,9 +162,13 @@ object ScalajsReact {
         "com.github.japgolly.nyaya" %%% "nyaya-gen"  % Ver.Nyaya % "test",
         "com.github.japgolly.nyaya" %%% "nyaya-test" % Ver.Nyaya % "test",
         monocleLib("macro") % "test"),
-      jsDependencies ++= Seq(
-        (ProvidedJS / "sampleReactComponent.js" dependsOn "react-dom.js") % Test, // for JS Component Type Test.
-        "org.webjars.bower" % "sizzle" % Ver.SizzleJs % Test / "sizzle.min.js" commonJSName "Sizzle"),
+      npmDependencies in Compile ++= Seq(
+        "react-addons-test-utils" -> Ver.ReactJs
```
japgolly: This isn't `provided` scope. Hmmm, you may hate me slightly for this but I wonder if I was wrong to ask you to make the other React dependencies provided scope :grimacing:. Sorry. I guess this will need some thought. A world with scalajs-bundler is different than the world before it.

Maybe it would be a good to idea to have mandatory npm deps declared with compile scope and transitive, and have the optional ones declared with provided scope. For me, it's a question of trust in scalajs-bundler. Let's say I publish this library with exported npm dependencies on React 15.3.2, if a downstream user declares the version to be 15.4.0 in their own project's npmDependencies, how confident are you that scalajs-bundler handle it correctly? What I mean by correctly is that after dependency resolution React 15.3.2 shouldn't be referenced at all, any transitive deps of 15.3.2 that aren't transitive deps of 15.4.0 shouldn't be referenced either, obviously version updates of React's transitive deps should be acknowledged rather than those of 15.3.2 being maintained. Dependency resolution with large graphs isn't easy, and in the NPM world all dep graphs are huge. How does scalajs-bundler handle all of that? And is it automatedly verifiable?

japgolly:
Hi @julienrf, I had another look through the commit and added more comments.
I also created a new branch and targeted this PR at it. Unfortunately it'll be another two weeks until I can get back to focusing on Scala.js stuff but once I can I'd like to delve into scalajs-bundler. It looks really good! :+1:

project/Build.scala
julienrf: Currently there is not such resolution mechanism (though it is planned, see https://github.com/scalacenter/scalajs-bundler/issues/55) so it is probably better to stick to “provided” dependencies.

julienrf: This `react-addons-test-utils` is only used to run scalajs-react tests, if I understand correctly, so it will not be transitively added to scalajs-react users projects.

fdietze:
Any news on this? I'd like to test and help.

julienrf:
For information, you don’t *have to* use scalajs-bundler in scalajs-react. You could just define the interface with `@JSImport` and let users choose whatever solution they want to effectively supply the react NPM module. You can have a look at how it has been done in sri.

japgolly:
Hey @fdietze & @julienrf, no update on my side. From memory I left a bunch of feedback for @julienrf but I don't think he followed up with changes. I suggest re-reading all the discussion in this PR and finishing off what's remaining, @fdietze.

Also, in reply to the last comment (@julienrf) that would likely be a very nice change if we can support both users and non-users of scalajs-bundler. It'd be worth exploring that approach.

julienrf:
@japgolly I might not have time to continue working on that, but here are suggestions for possible solutions.

- Define only one facade API, which uses `@JSImport`
  - Not tied to scalajs-bundler, but the facade would still expect to be used in a CommonJS like environment, so users would have to rely on a way to provide the React NPM modules they use, and I don’t think there are really a lot of alternatives than directly using scalajs-bundler for that. Said otherwise, this solution would not be tied to scalajs-bundler but would not be super convenient to use outside of scalajs-bundler ;
  - That’s the approach taken by sri, and examples require to both compile the project with sbt and then use npm to make it executable.
- Define two facade APIs, one relying on the global namespace (easy to use with `jsDependencies`) and one using `@JSImport` (easy to use with scalajs-bundler)
  - Requires a little more work but would give a lot of flexibility to the users
  - We would have the following abstract facade definition:
    ~~~ scala
    @js.native trait React extends js.Object { def createElement(…): ReactDOMElement = js.native }
    ~~~
  - Then we would have a first concrete facade like this:
    ~~~ scala
    @JSName("React") @js.native object React extends React
    ~~~
  - And, in another artifact or package, the following concrete facade:
    ~~~ scala
    @JSImport("react", Namespace) @js.native object React extends React
    ~~~

fdietze:
Sorry, I had to step back a bit since I struggled to get my D3 Facade to work with both bundler and jsDependencies. I'll look back into this pull request, when I touch my react project again...

japgolly:
@julienrf No worries and thank you for the work you've done thus far. The advice on what to do next is also very helpful. This will eventually get picked up and finished, if not by someone else, then by me. Thanks again!

@fdietze I understand & np; thanks for letting me know. :)

japgolly:
I don't see how I can include this feature and still be able to use scalajs-react in my own work project. :(

It seems (https://github.com/scala-js/scala-js/pull/2794#issuecomment-285250479) I'm faced with 3 unhappy choices:

1. Put `@JSName` on my facades and don't allow users to use `CommonJSModule`.
2. Put `@JSImport` on my facades and force users to use `CommonJSModule` and probably force them to use scalajs-bundler which will in turn force them to only import deps a certain way and, only output one huge single JS file.
3. Double build and double publish all of my library artifacts so that users can choose whether they want to use modules / script style.

Right now we're on option 1, a PR like this would take us to option 2 forcing unacceptable changes to my work-project's assets and  would negatively affect user experience - download parallelism, cache granularity. Option 3 would be hideous.

I don't see an acceptable way forward at the moment.
*Or* (hopefully), I'm mis-understanding how everything fits together...

japgolly:
Rebased here: https://github.com/japgolly/scalajs-react/commits/topic/scalajs-bundler

japgolly:
Closing as detailed in #343.

@julienrf thanks again for the PR. It's been rebased onto `topic/scalajs-bundler`, revised for 1.0.0-RC1, and will likely be revised again and merged when Scala.JS 1.0 comes out. Also it has some weird test failure with timeouts that wasn't there before so if you have time and are feeling generous it'd be fantastic if you could look at it. It *might* be a scalajs-bundler bug but I haven't confirmed that 100%. Cheers.

PR #323, opened 2016-12-11
Adding link to Scaladoc in README.md

jhegedus42:


japgolly:
First of all, AWESOME! I didn't know it was publicly, freely and automatically hosted.
Second, the reason I haven't merged this is because I want to add links to the ScalaDoc of each module, not just the `core` module. I think I didn't reply because I was thinking it would be easier to do it myself with formatting the way I want rather than ask you and then keep bike-shedding replies, making you (rightfully) annoyed. Anyway, ran out of time and forgot, won't have time until Feb. It's a bit menial but feel free to update the PR, else I'll leave this outstanding for another few weeks.

jhegedus42:
No rush, it's good to know that Scaladoc exists :)

japgolly:
Cool, merged and updated in 50cda2e03d567f01ef19280af172936d4ac06bc3. Thanks @jhegedus42

jhegedus42:
Happy to contribute if I can.

PR #328, opened 2017-02-06
Allow asynchronously rendering a React component into the body and document

jcrosetto:
This adds the ability to render React components into the body and document which then wait for a Future to complete before unmounting and doing cleanup. This works well with the Dsl.Future in [test-state](https://github.com/japgolly/test-state). It was the recommended improvement to address issue https://github.com/japgolly/test-state/issues/7.


japgolly:
Awesome! Thanks very much @jcrosetto!

PR #337, opened 2017-03-13
Make Callback.when lazy

Daxten:
hey,
Callback.when should be evaluated lazily in my opinion, one example:

```scala
DifferentComponent(callback = Callback.when(dom.confirm("Really?"))(doSomething))
```

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/Callback.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/Callback.scala
@@ -91,7 +91,7 @@ object Callback {
    * @param cond The condition required to be `true` for the callback to execute.
    */
   def when(cond: Boolean)(c: => Callback): Callback =
-    if (cond) c else Callback.empty
+    Callback.lazily(if (cond) c else Callback.empty)
 
   /**
    * Convenience for applying a condition to a callback, and returning `Callback.empty` when the condition is already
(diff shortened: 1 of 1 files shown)


Daxten:
I just saw how Callback.lazy is implemented and think my second commit would be a better fit for this.
I'm also interested in input if someone thinks this is a bad idea

japgolly:
Hey. The problem with that is that you're mixing pure/impure code. `dom.confirm("Really?")` is impure and should be wrapped in `CallbackTo` (makes it reusable too). What do you think about one of these instead?

```scala
val prompt = CallbackTo(dom.confirm("Really?"))
prompt.flatMap(Callback.when(_)(doSomething)

// or better:
val prompt = CallbackTo(dom.confirm("Really?"))
prompt.flatMap(doSomething.when)

// or if there are newbies on the team
for {
  allow <- CallbackTo(dom.confirm("Really?"))
  _     <- doSomething.when(allow)
} yield ()
```

Daxten:
yeah that looks better.. maybe there could be a 

```
def when(prop: CallbackTo[Boolean]): Callback
```

? Also after thinking about it my implementation wouldn't even fix my code :)

japgolly:
We could overload it....
```
def when(cond: Boolean): Callback
def when(cond: CallbackTo[Boolean]): Callback
```

PR #341, opened 2017-03-15
Support for cats

alonsodomin:
Took the liberty of writting an "extra" module with integration between ScalaJS-React and Cats. I did not port yet the `Fix` type in the state extension but it should be easy to add it. Just want to make this PR to start getting some feedback.

Any ideas on how to test this could would be grateful, I did not find any tests in the same module for Scalaz.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,6 +9,7 @@ val core          = ScalajsReact.core
 val extra         = ScalajsReact.extra
 val scalaz72      = ScalajsReact.scalaz72
 val monocle       = ScalajsReact.monocle
+val cats          = ScalajsReact.cats
 val test          = ScalajsReact.test
 val ghpagesMacros = ScalajsReact.ghpagesMacros
 val ghpages       = ScalajsReact.ghpages
--- a/cats/src/main/scala/japgolly/scalajs/react/CatsReact.scala
+++ b/cats/src/main/scala/japgolly/scalajs/react/CatsReact.scala
@@ -1,0 +1,10 @@
+package japgolly.scalajs.react
+
+import japgolly.scalajs.react.internal._
+
+/**
+  * Created by alonsodomin on 13/03/2017.
+  */
+trait CatsReact extends CatsReactExt with CatsReactInstances with CatsReactState
+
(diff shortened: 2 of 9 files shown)


japgolly:
Hi @alonsodomin, thanks for contributing! I had a quite skim and it looks fine. Is it just a copy-paste of Scalaz? If so I'm curious, what was the reason for holding off on `Fix`?

As for tests, all modules are tested together in `test/src/test`. The Scalaz module is tested directly in [`ScalazTest.scala`](https://github.com/japgolly/scalajs-react/blob/master/test/src/test/scala/japgolly/scalajs/react/ScalazTest.scala). It's not large but neither is the content of the `scalaz` module that isn't covered by parametricity. :)

alonsodomin:
Thanks for the pointer, I should have checked deeper in the `test` module. The inclusion of Cats is not a 100% copy and paste, i.e.: There in the ReactST monad I renamed the `getsXX` (`get`, `getsT` and `getsM`) by `inspectXX` methods, as that is how they are called in Cats VS Scalaz.

Arguabily even most of those methods could be written to require an instance of an `Applicative` instead of a `Monad` as Cats does not require the `M[_]` to be always one, whilst Scalaz does but I thought it was better to keep it the same way as in the Scalaz module.

So I will add the instances for `Fix` and some tests to have these more complete.

japgolly:
Ah right, I understand. That's a good eye for detail :)

If you want to weaken `Monad` to `Applicative` I think that's very appropriate. It'd only be an annoyance for someone transitioning their Cats-based `ReactS` to Scalaz which I doubt would ever happen and even if it did, different constraints are the reality of each library so I think it's good to reflect that.

alonsodomin:
Right, mostly everything has been ported. The most significative thing left out is all the support for the `IO` monad, as Cats does not have one and there is no purpose on adding one. This "forced" me to use the `Id` monad in the test side, which is pretty lame, but all least it verifies that the tests pass.

japgolly:
Looks great, let's merge it! Thanks again for the contribution.

PR #346, opened 2017-03-26
Add an additionaly addCondition options for rules that allows the rul…

bpatters:
Add an additional addCondition options for rules that allows the rule to take in a page as context for evaluation.
 This allows us to do things like: /#share/uuid  ---> if user is authed AND the current authed share uuid is equal to requested uuid then render page, otherwise evaluate unmet..

Basic idea here is to make the Page Context available to the condition evaluator.


japgolly:
Great idea! Thanks

Issue #354, opened 2017-04-26
Bump Scala/ScalaJS versions for Scalajs-react v0.11.x


mdedetrich:
Since 0.6.15 is the last version for scalajs 0.6.x series (which adds a lot of improvements) and due to how scalajs binary compatibility works, I created a branch which bumps the versions in the 0.11.x series of scalajs-react.

The intention is that this should be the last release of scalajs-react for the 0.11.x series and since there are a lot of projects with non trivial code sizes that use scalajs-react 0.11.3, in the interests of binary compatibility it would be good to have a release which uses the latest version of scalajs

I have implemented the changes in https://github.com/mdedetrich/scalajs-react/tree/v0.11.4. The only other possible change that I can see is bumping the base react version from 15.3.2, however I think that is out of scope and will cause more problems.

@japgolly Would it be possible to create a branch based off the v0.11.3 tag (https://github.com/japgolly/scalajs-react/tree/v0.11.3), probably called v0.11.x, so I can make a PR against this branch, and then release 0.11.4?

Also I think in general its a good idea to have a 0.11.x branch for scalajs-react if issues like this prop up in the future (and we can trivially bump versions)

japgolly:
Sure, all very good points, I'll create a `series/v0.11.x` branch now. :+1:

mdedetrich:
Scalajs 0.6.16 just came out now, let me quickly rebump and make another PR

japgolly:
Alrighty v0.11.4 is on its way to maven central

mdedetrich:
Ill close the issue then, thanks a lot!

PR #356, opened 2017-04-30
v0.11.4 with scala.js 0.6.16

mdedetrich:
This is the PR to address https://github.com/japgolly/scalajs-react/issues/354. Idea is to release a version of the scalajs-react 0.11.x series with the latest dependencies for scala.js (in this case its scala.js 0.6.16 which is the latest compatible version of Scalajs in the 0.6.x series)

Note that scala.js 0.6.16 is both backwards and forwards compatible with 0.6.15, hence why 0.6.15+ is the requirement, see https://www.scala-js.org/news/2017/04/29/announcing-scalajs-0.6.16/ for more info

@japgolly Does it make sense to bump JSDom as well? According to the release notes,

> It also adds support for jsdom v10.x (which contains breaking changes wrt. jsdom v9.x) and Scala 2.13.0-M1.

This is probably the main change for scala.js 0.6.16 vs 0.6.15

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: scala
 sudo: false
 
 scala:
-  - 2.11.8
-  - 2.12.0
+  - 2.11.11
+  - 2.12.2
 
 jdk:
   - oraclejdk8
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Includes a router, testing utils, performance utils, more.
   - [Performance Management](doc/PERFORMANCE.md).
   - [Smaller stuff](doc/EXTRA.md).
 - [Testing](doc/TESTING.md).
-- [Changelogs](doc/changelog) — [Latest](doc/changelog/0.11.3.md).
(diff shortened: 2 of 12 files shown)


mdedetrich:
@japgolly Also this build is failing due to
```
The command "bin/checkDevOnly" exited with 1.
```

Is this intended?

japgolly:
Awesome, LGTM and thanks for adding the changelog as well. :+1: 

> Does it make sense to bump JSDom as well

If you mean the scala-js-dom, that's already at the latest version. When it means in the 0.6.16 notes is compatibility with the npm library [jsdom](https://github.com/tmpvar/jsdom) which in our case means it only affects Travis and by affects I mean it now works again where as it stopped working once jsdom v10 was released.

> this build is failing due to `The command "bin/checkDevOnly" exited with 1.`

There seems to be a problem with 0.6.16 that causes that but there's a workaround available on master. I'll cherry-pick it after merging this PR.

PR #360, opened 2017-05-03
Add a note about evaluation semantics of conditional rendering sugar.

vpavkin:
Please, correct me if I'm wrong somewhere...


japgolly:
Aye, this is true.

PR #361, opened 2017-05-04
Integrate with Cats' IO

alonsodomin:
Finally, Cats it's getting is own `IO` monad and apparently it's the best thing after sliced bread since it comes with effect-related typeclasses and laws. In relation with SJS-ReactJS this means that the Cats integration module can now provide the same integration level than the Scalaz one.

One important remark, Cat's `IO` is closer to `scalaz.Task` (and `fs2.Task` / `monix.Task`) than Scalaz's `IO`. What this means is that a `cats.effect.IO[A]` may represent an asynchronous computation in which the operation `unsafeRunSync()` will throw an Exception ScalaJS since blocking threads is not possible in JavaScript.

The solution for the previous problem is achieved by consuming the real `IO` asynchronously via `runAsync` - which will convert any `IO[A]` into a `IO[Unit]` - and then run synchronously this very last one. Something in the likes of:

```
val myIO: IO[A] = ...
val handler: Either[Throwable, A] => IO[Unit] = ...
myIO.runAsync(handler).unsafeRunSync()
```

Unfortunately it's not possible to actually do this everywhere we need to do an `.unsafeRunSync()` so I'm just leaving it as it is, open for discussion whether SJS-ReactJS should even care about this.

### DIFF
--- a/cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactExt.scala
+++ b/cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactExt.scala
@@ -1,6 +1,7 @@
 package japgolly.scalajs.react.internal
 
 import cats.{Eq, Monad, ~>}
+import cats.effect.IO
 
 import japgolly.scalajs.react._
 import japgolly.scalajs.react.extra._
@@ -9,6 +10,15 @@ import japgolly.scalajs.react.extra._
   * Created by alonsodomin on 13/03/2017.
   */
 object CatsReactExt {
+
+  final class CallbackToOps[A](private val body: () => A) extends AnyVal {
+    private def callback = CallbackTo.lift(body)
+
+    def toIO: IO[A] = IO(body())
+
+    def flattenIO[B](implicit ev: A =:= IO[B]): CallbackTo[B] =
+      callback.map(_.unsafeRunSync())
+  }
 
   final class MA[M[_], A](private val ma: M[A]) extends AnyVal {
     def toCallback(implicit t: M ~> CallbackTo): CallbackTo[A] = t(ma)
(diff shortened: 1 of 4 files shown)


japgolly:
So is it 'unsafeRunSync` that may or may now throw, and `unsafeRunAsync` that always works, is that right? (I'm tired and please loosely interpret this until it makes sense.) Using promises we can turn all instances of `IO[A]` into `Future[A]` without throwing (unless the underlying code throws). Considering that we don't need to match the Scalaz extensions exactly -- none of this is mandatory -- I think it'd be better to either omit the stuff that may throw (`.unsafeRunSync`) or change it so that output `A`s are `Future[A]`s instead.

alonsodomin:
Yeah, it's like that, since users can create async `IO`s using `IO.async` (similar to Scalaz's `Task.async`), the method `unsafeRunSync` may need to block to be able to get the `A` out of the `IO[A]` and since in ScalaJS you can not block...

There is a `unsafeToFuture` in Cats' IO that basically represents a `IO[A] ~> Future[A]` which it's created using a promise. So when integrating this with callbacks, the transformations could be `IO[A] ~> CallbackTo[Future[A]]` instead...

japgolly:
If you want to go down that road, I'm fine with it. :)

alonsodomin:
I've been playing again with this PR and I think it might not be worthy as it is not possible to get fully safe implementation. Even when making the natural transformation `IO ~> CallbackTo` a `IO ~> CallbackTo[Future[?]]` (which forces a dance of type signatures in the state related functions to be able to resolve the implicit), then when implementing `Effect`, the `extract` operation is only possible by using `IO.unsafeRunSync()`, which is unsafe because of its blocking feature.

Probably there is a path to explore by following a transformer approach with a base `CallbackToT[M[_], A](val runNow: () => M[A])` and a `MonadCallback` typeclass, such that in Cats `IO ~> CallbackToT[Future, ?]`. But this will either force rewritting monad instances in the core or tie it one of the FP libraries, which it´s not desired in the library's core.

So, probably it's better to not even attempt this, close this PR, and force users to integrate manually in a case by case basis.

japgolly:
Yeah no worries. I definitely don't want to tie core to any specific FP lib. Feel free to revisit later

japgolly:
Oh and sorry for the delay. I looked at this the other day but was too sleep-deprived to process it. :)

PR #367, opened 2017-05-15
Fixing link to live template.

ikamthania:


japgolly:
Ah, I misread the UI, you're right, thanks

PR #368, opened 2017-05-19
Expand monocle ext dependency instructions

mgilbertnz:
Hi there,

Thanks for writing this library.  It kicks ass.  You're blog is interesting too!

One thing I got stumped on for quite a while was trying to work out why my `StateSnapshot` was not compiling when I tried to use Monocle lenses.  After some time I worked out that I was missing the macro paradise compiler plugin.  I thought I'd throw in a pull request to help others who make the same mistake.

I've expanded the necessary instructions in the docs to include this.  Feel free to comment and change.

Cheers,
Matt

### DIFF
--- a/doc/FP.md
+++ b/doc/FP.md
@@ -39,7 +39,13 @@ Monocle
 =======
 
 ```scala
-libraryDependencies += "com.github.japgolly.scalajs-react" %%% "ext-monocle" % "1.0.0"
+libraryDependencies ++= Seq(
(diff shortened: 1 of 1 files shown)


japgolly:
Hey Matt, cheers for the feedback! Awesome to hear that you're enjoying it!

Thanks for this PR. A part of me thinks that if one is already using Monocle then the Monocle docs should have the most up-to-date info, there's a danger this could go out-of-date. ...Which when I think about it is such a trivial downside and it's much nicer to include the extra info and  make it clear & easy for new users. Good call. Merging and thanks again!

PR #369, opened 2017-05-20
Fix reusable functions example

mgilbertnz:
Couple of quick doc corrections.

Were trying to pass a `Long` to `withKey` which isn't able to be converted into a `Key`.

`withKey()` takes `Key` which is:
`type Key = String | Boolean | JsNumber | Null`
`type JsNumber = Byte | Short | Int | Float | Double`

### DIFF
--- a/doc/PERFORMANCE.md
+++ b/doc/PERFORMANCE.md
@@ -213,7 +213,7 @@ type PersonId = Long
 type PersonData = String
 
 val topComponent = ScalaComponent.builder[State]("Demo")
-  .initialState_P(identity)
+  .initialStateFromProps(identity)
   .renderBackend[Backend]
   .build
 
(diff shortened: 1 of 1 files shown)


doc/PERFORMANCE.md
```diff
@@ -213,7 +213,7 @@ type PersonId = Long
 type PersonData = String
 
 val topComponent = ScalaComponent.builder[State]("Demo")
-  .initialState_P(identity)
+  .initialStateFromProps(identity)
```
japgolly: Ah dammit, how did I miss this and what else have I missed?...

japgolly:
Let me double-check that the implementation is correct. AFAIR Scala.JS doesn't represent `Long`s as JS numbers...

japgolly:
Thanks for the PR

mgilbertnz:
"AFAIR Scala.JS doesn't represent Longs as JS numbers..." - Yeah, hence why I opted for the `.toString` an alternative would be to modify the example to use a Double or Int rather than a Long.

I've just spent the last week studying this library so I'll send through any more corrections I spot.

japgolly:
> I've just spent the last week studying this library so I'll send through any more corrections I spot.

Much appreciated!

PR #383, opened 2017-06-08
Escape vertical bar

cquiroz:
Just a small fix for ROUTER documentation


japgolly:
Hmm, that's new. Wish Github would stop changing their damn parser. Oh wells good to know. Thanks for the fix

PR #389, opened 2017-06-15
Add references to diode and scalafiddle-editor in README

malaman:
Hi, 

thanks for package!
It is really cool, that so solid ecosystem already exists around scalajs-react.

For me [diode](https://github.com/suzaku-io/diode) and [scalafiddle-editor](https://github.com/scalafiddle/scalafiddle-editor) projects were very helpful for understanding how to build frontend apps with scalajs-react.
 
I've added references to the packages in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,11 @@ Includes a router, testing utils, performance utils, more.
   * [scalajs-benchmark](https://github.com/japgolly/scalajs-benchmark/)
   * [chandu0101 / scalajs-react-components](https://github.com/chandu0101/scalajs-react-components)
   * [payalabs / scalajs-react-mdl](https://github.com/payalabs/scalajs-react-mdl) - (Material Design Lite components)
+  * [diode](https://github.com/suzaku-io/diode) - library for managing application state, influenced by Flux and Elm
+
+* Open Source Projects, which are using [scalajs-react](https://github.com/japgolly/scalajs-react)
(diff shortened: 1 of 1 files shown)


japgolly:
Good idea, thanks! One little nit pick about Diode:

> library for managing immutable application state with unidirectional data flow

I'm not sure if just me but it seems to imply that unidirectional data flow isn't supported/possible out-of-the-box, where as it is. I think the more defining attribute Elm-style architectures is that the model/flow is global. As far as unidirectional data flow, React provides that out of the box in the micro, and I achieve it in the macro in a modular, composable fashion (as opposed to global) using optics and the `ext-monocle` scalajs-react module. Should we tweak the description? What do you think?

malaman:
Thanks for feedback.
I will update PR tomorrow and remove unidirectional data flow texting from diode description.
Also I will add scalacenter/scastie to list of open source projects, which are using scalajs-react.

Global state management (flux/redux) is very popular in JS based React projects and I believe what people, which come from js side (like me) automatically start to look for something similar in scala.js.

Other ways of state management is not very known. 
I will look in ext-monocle approach and try to create another PR with the doc regarding possible options for state management.

malaman:
PR is updated.

japgolly:
LGTM, thanks a lot for this - it's a good idea to link to more of the ecosystem - and sorry about the delay

PR #393, opened 2017-06-22
Change remaining SVG attrs to React SVG attrs

mgilbertnz:
Been working with SVG recently and I've noticed a lot of warnings such as
`Warning: Unknown DOM property stroke-linejoin. Did you mean strokeLinejoin?`

This appears to be due to SvgAttrs being half-migrated.

Looking at the React source, it appears that the remaining "-" attrs are now supported using React camel case attrs.
https://github.com/facebook/react/blob/master/src/renderers/dom/shared/SVGDOMPropertyConfig.js

This pull request moves the remaining attrs to camel case to avoid the above warning.


japgolly:
I'm surprised that went unnoticed for so long. Thanks!

PR #394, opened 2017-06-23
Add runStateFuture and runStateFnFuture functions

cfraz89:
I may be in over my depth, but here goes.

I would like to use the ReactS monad along with ajax requests. To accomodate this, I've added some convenience functions into the Cats StateAccessRW extension.

These allow the monad to be used as such:
```scala
object Example {
  case class State(list: List[String])

  val ST = ReactS.Fix[State]

  val home = ScalaComponent.builder[Unit]("Example")
    .initialState(State(Nil))
    .renderS { ($, s) =>
      <.div(
        ^.onClick --> $.runStateFuture(appendExample),
        <.ul(s.events.toTagMod(<.li(_)))
      )
    }
      .componentDidMount(_.runStateFuture(appendExample))
    .build

  def appendExample: ReactST[Future, Home.State, Unit] =
    ST.modM(s => Ajax.get("http://date.jsontest.com/") map (r => State(s.events :+ r.responseText)))
}
```

Hopefully there is nothing fundamentally wrong with using the monad in such a fashion.

### DIFF
--- a/cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactState.scala
+++ b/cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactState.scala
@@ -3,10 +3,9 @@ package japgolly.scalajs.react.internal
 import cats._
 import cats.data.StateT
 import cats.implicits._
-import japgolly.scalajs.react._
-import CatsReact.{CatsReactExt_ReactST, reactCallbackCatsInstance}
 
-import scala.concurrent.{ExecutionContext, Future}
+import japgolly.scalajs.react._
+import CatsReact.{reactCallbackCatsInstance, CatsReactExt_ReactST}
 
 /**
   * Created by alonsodomin on 13/03/2017.
@@ -44,7 +43,6 @@ object CatsReactState {
 
   type ReactST[M[_], S, A] = StateT[M, StateAndCallbacks[S], A]
   type ReactS[S, A] = ReactST[Id, S, A]
-  final type TransAsync[M[_], N[_]] = M ~> ({type X[a] = CallbackTo[N[a]]})#X
 
   object ReactS {
     final def StateAndCallbacks[S](state: S, cb: Callback = Callback.empty) =
@@ -204,18 +202,18 @@ object CatsReactState {
   final class Ext_StateAccessRW[F[_], SI, S, Out[_]](si: SI)(implicit sa: StateAccessor.ReadWrite[SI, F, F, S], fToCb: Effect.Trans[F, CallbackTo], cbToOut: Effect.Trans[CallbackTo, Out]) {
     implicit private def autoOut[A](a: CallbackTo[A]): Out[A] = cbToOut(a)
 
+    private type TransAsync[M[_], N[_]] = M ~> ({type X[a] = CallbackTo[N[a]]})#X
+
     private def stateCB: CallbackTo[S] = fToCb(sa.state(si))
 
     private def run[M[_], A, B](st: => ReactST[M, S, A], f: (S, S, A, => Callback) => CallbackTo[B])(implicit M: M ~> CallbackTo, N: Monad[M]): CallbackTo[B] =
       stateCB.flatMap { s1 =>
         val runCB: CallbackTo[(StateAndCallbacks[S], A)] = M(st run StateAndCallbacks(s1))
         runCB.flatMap { x2 =>
-          val s2: StateAndCallbacks[S] = x2._1
-          val a: A = x2._2
-
-          def cb: Callback = fToCb(sa.setStateCB(si)(s2.state, s2.cb))
-
-          val res: CallbackTo[B] = f(s1, s2.state, a, cb)
+          val s2 : StateAndCallbacks[S] = x2._1
+          val a  : A                    = x2._2
+          def cb : Callback             = fToCb(sa.setStateCB(si)(s2.state, s2.cb))
+          val res: CallbackTo[B]        = f(s1, s2.state, a, cb)
           res
         }
       }
--- a/cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactState.scala
+++ b/cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactState.scala
@@ -7,6 +7,8 @@ import cats.implicits._
 import japgolly.scalajs.react._
 import CatsReact.{reactCallbackCatsInstance, CatsReactExt_ReactST}
 
+import scala.concurrent.{ExecutionContext, Future}
+
 /**
   * Created by alonsodomin on 13/03/2017.
   */
@@ -230,6 +232,15 @@ object CatsReactState {
 
     def runStateFnF[I, M[_], A](f: I => ReactST[M, S, A])(implicit M: M ~> CallbackTo, N: Monad[M], F: ChangeFilter[S]): I => Out[A] =
       i => runStateF(f(i))
+
+    def runStateFuture(st: => ReactST[Future, S, Unit])(implicit F: FlatMap[Future], ec: ExecutionContext): Callback =
+      stateCB.flatMap { s1 =>
+        val runCB = st.runS(ReactS.StateAndCallbacks(s1))
+        Callback.future(runCB.map { s2 => fToCb(sa.setStateCB(si)(s2.state, s2.cb)) })
+      }
+
+    def runStateFnFuture[I](f: I => ReactST[Future, S, Unit])(implicit F: FlatMap[Future], ec: ExecutionContext): I => Callback =
+      i => runStateFuture(f(i))
 
     def modStateF(f: S => S, cb: Callback = Callback.empty)(implicit F: ChangeFilter[S]): Out[Unit] =
       stateCB.flatMap(s1 =>
--- a/cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactState.scala
+++ b/cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactState.scala
@@ -232,13 +232,13 @@ object CatsReactState {
     def runStateFnF[I, M[_], A](f: I => ReactST[M, S, A])(implicit M: M ~> CallbackTo, N: Monad[M], F: ChangeFilter[S]): I => Out[A] =
       i => runStateF(f(i))
 
-    def runStateAsync[M[_]](st: => ReactST[M, S, Unit])(implicit F: FlatMap[M], C: M[Callback]=>Callback, ec: ExecutionContext): Callback =
+    def runStateAsync[M[_], A](st: => ReactST[M, S, A])(implicit F: FlatMap[M], C: M[Callback]=>Callback, ec: ExecutionContext): Callback =
       stateCB.flatMap { s1 =>
         val runCB = st.runS(ReactS.StateAndCallbacks(s1))
         C(F.map(runCB)(s2 => fToCb(sa.setStateCB(si)(s2.state, s2.cb))))
       }
 
-    def runStateFnAsync[M[_], I](f: I => ReactST[M, S, Unit])(implicit F: FlatMap[M], C: M[Callback]=>Callback, ec: ExecutionContext): I => Callback =
+    def runStateFnAsync[I, M[_], A](f: I => ReactST[M, S, A])(implicit F: FlatMap[M], C: M[Callback]=>Callback, ec: ExecutionContext): I => Callback =
       i => runStateAsync(f(i))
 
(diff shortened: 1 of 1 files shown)


japgolly:
Hi @cfraz89. The methods you're adding a bit too specific so I started simplifying the signatures to get to the root of what we're really doing here:

```scala
def runStateFuture(st: => ReactST[Future, S, Unit])(implicit F: FlatMap[Future], ec: ExecutionContext): Callback
def runStateFuture(st: => ReactST[Future, S, Unit]): Callback
def runStateFuture[A](st: => ReactST[Future, S, A]): CallbackTo[A]
def runStateFuture[M[_], A](st: => ReactST[M, S, A]): CallbackTo[A]
def runStateFuture[M[_], A](st: => ReactST[M, S, A]): Out[A] // From Ext_StateAccessRW
```
which is exactly the same (excluding implicit args) as the existing `runState` method.
```scala
def runState[M[_], A](st: => ReactST[M, S, A]): Out[A]
```

which matched my intuition. The existing methods should be generic enough to handle your case, you'll just need to provide the right implicits. The full signature is
```scala
def runState[M[_], A](st: => ReactST[M, S, A])(implicit M: M ~> CallbackTo, N: Monad[M]): Out[A]
```
so if you provide an implicit `Future ~> CallbackTo` and `Monad[Future]` it will work.... But as you probably guessed those will be unlawful and dangerous because `Future` itself is unlawful and dangerous. I'd recommend using another type similar to Scalaz `Task` or Monix's `Task`. Not sure what cats has in this space.

cfraz89:
Hi, I did initially go down that route, however doing it requires providing an implementation of Future[A] -> CallbackTo[A] ( to satisfy FlatMap[A]), which as far as I can see is impossible?

japgolly:
Yeah that's right, `CallbackTo` in synchronous although it does have some async methods on it:
https://github.com/japgolly/scalajs-react/blob/7ab69eb92a8e2afc595fa60b6057d7a50368c1ea/core/src/main/scala/japgolly/scalajs/react/Callback.scala#L595-L625

So while `CallbackTo[A]` is sync and safe, `CallbackTo[Future[A]]` is async and safe so long as the futures are created from within the callback. If you don't care about the results you can just call `.void` and it becomes a `Callback` again.

For example:
```scala
CallbackTo(something()) // CallbackTo[A]
  .async  // CallbackTo[Future[A]]
  .void  // Callback
```

With the exception of a value in a side-effect capturing type, using `Future` in anywhere (even in `ReactST` or `StateT` itself) is going to be dangerous and untrustworthy. `ReactST[Future` is already in the danger-zone. The usual approach is to wrap the Future in something safer or use a safe alternative like `Task`.
Does that help or am I missing the mark?

cfraz89:
Yep, however there is still the dillema of how to use the state monad with any form of asynchronous context:

given any async monad X (eg Future, Task, Cats IO etc)

def state: ReactST[X, State, A] =
    ST.setM(s => (get X) map (r => State(r)))

runState depends upon an implementation of X[A] ~> Callback[A].
This will always be impossible without blocking no?

cfraz89:
Here is what it would look like abstracted out to generalised async invocation (in the form of an implicit suguar class)

```scala
  implicit class StateRWFutureExt[P, S, B](stateRW: StateRW[P, S, B]) {

    def runStateAsync[F[_]](st: => ReactST[F, S, Unit])(implicit toCallback: F[Callback]=>Callback, FM: FlatMap[F]): Callback =
      toCallback(FM.map(st runS (ReactS.StateAndCallbacks((stateRW.state)))) (s => stateRW.setState(s.state, s.cb)))

    def runStateFnAsync[F[_], I](f: I => ReactST[F, S, Unit])(implicit toCallback: F[Callback]=>Callback, FM: FlatMap[F]): I => Callback =
      i => runStateAsync(f(i))

  }
```

japgolly:
This looks safe but it's not:
```scala
implicit def futureToCallback(f: Future[Callback])(implicit ec: ExecutionContext): Callback = Callback.future(f)
```
or else I would've like to add a similar conversion. `Callback`s are pure and repeatable where as by the time you have an instance of `Future` the horses have already bolted and the barn's already on fire. The types lie because `Future` lies.

I see what you're saying about the `M ~> CallbackTo` preventing async workloads...
`M[A] => CallbackTo[A]` is impossible if `M` is async. What do you think about `M[A] => CallbackTo[F[A]]` instead where `F` would typically be `Future`? That way you could build up your state monad using a safe, pure async effect type like `Task` (or even `Future` directly if you like to live dangerously and accept the void warranty), and then run it in a safe context (`CallbackTo`) to return a `Future` which is easy to work with.

cfraz89:
Sounds good, though I couldn't come up with an implementation for runStateAsync leveraging such a function. You'll probably be better at that 😅

cfraz89:
Unless A in your case would typically be the state monad itself? And not the A parameter as used by the runState* functions?

cfraz89:
Is something along these lines what you were picturing?

```scala

type CallbackF[A] = CallbackTo[Future[A]]

def runStateAsync[M[_], A](st: => ReactST[M, S, A])(implicit F: Monad[M], M: M ~> CallbackF, ec: ExecutionContext): CallbackF[A] = {
     stateCB.flatMap { s1 =>
         val runCB = st.run(ReactS.StateAndCallbacks(s1))
         M(runCB).map { f =>
             f.map { x2 =>
                 val s2: StateAndCallbacks[S] = x2._1
                 val a: A = x2._2

                 def cb: Callback = fToCb(sa.setStateCB(si)(s2.state, s2.cb))

                 cb.map(_ => a).runNow()
            }
        }
    }
}
```

cfraz89:
Ok think I've got what you were thinking of now...

```scala
def runStateAsync[M[_], A](st: => ReactST[M, S, A])(implicit N: Monad[M], ec: ExecutionContext): CallbackTo[M[A]] = {
      stateCB.map { s1 =>
        val runCB = st.run(ReactS.StateAndCallbacks(s1))
        runCB.map { x2 =>
            val s2: StateAndCallbacks[S] = x2._1
            val a: A = x2._2

            fToCb(sa.setStateCB(si)(s2.state, s2.cb)).runNow()
            a
        }
      }
    }
```

japgolly:
Hey @cfraz89, I was hoping to get back to you earlier but wasn't able to. Now something else big has come up and so it's going to be a while longer until I can give this attention. Sorry about that :(

japgolly:
Ok I finally had some time for this. In your two snippets above, as you know the first is hardcoded to `Future`, the second is better but there are two problems with it: 1) the output is fixed to `M` which is the same as the input, I'm not sure how useful that would be, maybe fine but it seems inflexible. 2) there's an escaping `.runNow()` which is unsafe.

I was thinking more along these lines:

```scala
  type TransAsync[M[_], N[_]] = M ~> ({type L[a] = CallbackTo[N[a]]})#L

  private def runAsync[M[_], N[_], A, B](st: => ReactST[M, S, A],
                                         f: (S, S, A, => Callback) => CallbackTo[B])
                                        (implicit M: Monad[M], N: Functor[N], trans: TransAsync[M, N]): CallbackTo[N[B]] =
    stateCB.flatMap { s1 =>
      type CN[X] = CallbackTo[N[X]]
      val runCB: CN[(StateAndCallbacks[S], A)] = trans(st run StateAndCallbacks(s1))
      runCB.flatMap[N[B]](nxa =>
        CallbackTo(
          N.map(nxa) { x2 =>
            val s2   : StateAndCallbacks[S] = x2._1
            val a    : A                    = x2._2
            def cb   : Callback             = fToCb(sa.setStateCB(si)(s2.state, s2.cb))
            val resCB: CallbackTo[B]        = f(s1, s2.state, a, cb)
            resCB.runNow() // Safe because surrounded by CallbackTo
          }
        )
      )
    }

  def runStateAsync[M[_], N[_], A](st: => ReactST[M, S, A])(implicit M: Monad[M], N: Functor[N], trans: TransAsync[M, N]): Out[N[A]] =
    runAsync[M, N, A, A](st, (s1, s2, a, cb) => cb.map(_ => a))
```

Here I too use a `.runNow()` but it's wrapped in `CallbackTo` which is safe.

Then outside of the library, to use it, you would need an implicit async transformation like:
```scala
  import scala.concurrent._
  import scala.concurrent.ExecutionContext.Implicits.global

  // Obviously you'd replace Function0 with something better
  implicit object Example extends TransAsync[Function0, Future] {
    def apply[A](f: Function0[A]): CallbackTo[Future[A]] =
      CallbackTo(Future(f()))
  }
```
The great parts here is that the `Future` creation is inside of `CallbackTo` making it actually safe and repeatable because its the constructor that side-effects.

WDYT?

cfraz89:
Aye looks very good to me. The ```scala ({type L[a] = CallbackTo[N[a]]})#L ```
My only question is - are the synchronous runState functions still required at this point? As far as I can tell they are functionally equivalent where N and M are Id.

Grogs:
@cfraz89 that's a type lambda. The [kind-projector plugin](https://github.com/non/kind-projector) gives it nicer syntax.

cfraz89:
Hey @japgolly,
I've implemented your suggestions.
Works for my use case, along with some auxilliary code to make it useful for cats io:

```scala
    def runStateIO[A](st: => ReactST[IO, S, A]): Callback =
      stateRW.runStateAsync[IO, IO, A](st) map (_.unsafeRunAsync(_ => ()))
    }

  val transIO = new CatsReactState.TransAsync[IO, IO] {
    override def apply[A](fa: IO[A]): CallbackTo[IO[A]] = CallbackTo(fa)
  }
```

Something about requiring TransAsync[IO, IO] seems off to me, but I can't really see an alternative.
If this is fine, maybe its something that could be integrated into [https://github.com/japgolly/scalajs-react/pull/361](url) at a later stage?

japgolly:
Hi @cfraz89, I'm sorry about the lack of response on this PR. I should be in a better position to give this more attention now.

I think this PR still needs work until I can include it. We can do this two ways. I can merge what you've done so far and then tinker around with it and make further changes myself (which is much faster for me but if you enjoy this it deprives you). Or else I can keep writing my suggestions and asking you to make changes -- that's ok with me, as this is your PR, but just be warned that you'll probably get hit with a lot of requests for change and it might take a while. I'm happy to do either so let me know.

Regardless of which option though, there are two changes you'll need to make first please. I'll comment inline.

cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactState.scala
```diff
@@ -3,9 +3,10 @@ package japgolly.scalajs.react.internal
 import cats._
 import cats.data.StateT
 import cats.implicits._
-
 import japgolly.scalajs.react._
-import CatsReact.{reactCallbackCatsInstance, CatsReactExt_ReactST}
+import CatsReact.{CatsReactExt_ReactST, reactCallbackCatsInstance}
+
+import scala.concurrent.{ExecutionContext, Future}
```
japgolly: None of this change is required

cats/src/main/scala/japgolly/scalajs/react/internal/CatsReactState.scala
```diff
@@ -208,10 +210,12 @@ object CatsReactState {
...
-          val s2 : StateAndCallbacks[S] = x2._1
-          val a  : A                    = x2._2
-          def cb : Callback             = fToCb(sa.setStateCB(si)(s2.state, s2.cb))
-          val res: CallbackTo[B]        = f(s1, s2.state, a, cb)
+          val s2: StateAndCallbacks[S] = x2._1
+          val a: A = x2._2
+
+          def cb: Callback = fToCb(sa.setStateCB(si)(s2.state, s2.cb))
+
+          val res: CallbackTo[B] = f(s1, s2.state, a, cb)
```
japgolly: I don't see any functional change here, just changes to formatting (and degradation at that IMO). Please restore.

cfraz89:
Hey japgolly, I'm ok with you taking over this, if you are of course. I've undone the stray reformatting as per your requests.

japgolly:
Hi @cfraz89. I've just merged this and I might do some refactoring to other methods in the area to use it to. 

I'm sorry that this ended up taking so much time to get through; you have the misfortune of having the worst experience of any PRs that've been merged! But it's in now and will go out in the next release. Thank you for the contribution.

japgolly:
Hi again @cfraz89, I made a few changes including renaming to `runStateM` because asynchronicity is user-provided and not innate. I also got rid of that pesky `runNow`. The code would've been a little nicer if Cats had a Distributive type-class (dual of Traverse) but no big deal.

cfraz89:
Hey, no problems. We got there :D

PR #399, opened 2017-07-20
gh-pages examples update

malaman:
Hi,

i've made some changes to gh-pages examples:
 - update js examples code (using functional components, classes and jsx syntax)
 - change ScalaComponent to ScalaFnComponent in some places

### DIFF
--- a/gh-pages/src/main/scala/ghpages/examples/ProductTableExample.scala
+++ b/gh-pages/src/main/scala/ghpages/examples/ProductTableExample.scala
@@ -13,121 +13,134 @@ object ProductTableExample {
 
   val jsSource =
     """
-      |var ProductCategoryRow = React.createClass({
-      |    render: function() {
-      |        return (<tr><th colSpan="2">{this.props.category}</th></tr>);
-      |    }
-      |});
+      |class ProductCategoryRow extends React.Component {
       |
-      |var ProductRow = React.createClass({
-      |    render: function() {
-      |        var name = this.props.product.stocked ?
-      |            this.props.product.name :
-      |            <span style={{color: 'red'}}>
-      |                {this.props.product.name}
-      |            </span>;
-      |        return (
-      |            <tr>
-      |                <td>{name}</td>
-      |                <td>{this.props.product.price}</td>
-      |            </tr>
-      |        );
-      |    }
-      |});
+      |  render() {
+      |    return (
+      |      React.createElement('tr', null,
+      |        React.createElement('td', {style: {fontWeight: 'bold'}}, this.props.category)
+      |      )
+      |    );
+      |  }
+      |}
       |
-      |var ProductTable = React.createClass({
-      |    render: function() {
-      |        console.log(this.props);
-      |        var rows = [];
-      |        var lastCategory = null;
-      |        this.props.products.forEach(function(product) {
-      |            if (product.name.indexOf(this.props.filterText) === -1 || (!product.stocked && this.props.inStockOnly)) {
-      |                return;
-      |            }
-      |            if (product.category !== lastCategory) {
-      |                rows.push(<ProductCategoryRow category={product.category} key={product.category} />);
-      |            }
-      |            rows.push(<ProductRow product={product} key={product.name} />);
-      |            lastCategory = product.category;
-      |        }.bind(this));
-      |        return (
-      |            <table>
-      |                <thead>
-      |                    <tr>
-      |                        <th>Name</th>
-      |                        <th>Price</th>
-      |                    </tr>
-      |                </thead>
-      |                <tbody>{rows}</tbody>
-      |            </table>
-      |        );
-      |    }
-      |});
+      |class ProductRow extends React.Component {
+      |  render() {
+      |    var name = this.props.product.stocked ?
+      |      this.props.product.name :
+      |      <span style={{color: 'red'}}>
+      |        {this.props.product.name}
+      |      </span>;
+      |    return (
+      |      React.createElement('tr', null,
+      |        React.createElement('td', null, name),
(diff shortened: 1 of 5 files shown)


malaman:
@japgolly should this PR be canceled (i've opened it because of deprecation of createClass in React.js library)?

japgolly:
Hi @malaman. No I'm still planning on getting to this. From today onwards I will finally have time in my week for OSS again, so I'll get to it soon. Thanks for your patience.

japgolly:
Firstly, thanks again for your patience and your work in creating and submitting this PR in the first place. It's always of great value to see PRs come in addressing the demos and documentation.

Unfortunately, there are reasons I didn't just hit the merge button ages ago: I'm not sure about some of the changes here...

#### 1. JSX

Updating the JS snippets to whatever is on the React site these days I great but I'm very torn as to whether we should be showing JSX. If one were to build a little button to toggle the JS snippets between JS and JSX then it'd be no problem but to show JSX and not JS bothers me. I think JSX makes more sense only to already experienced with raw React, where as a big (the biggest?) chunk of the site's audience is going to be Scala devs. To such an audience JSX doesn't make sense because it obfuscates the relationship to JS which is often what one would seek to understand when considering Scala.JS.

So I think either show JS instead of JSX, *or* add little JS↔JSX toggle buttons.

#### 2. ScalaFnComponent

I'm not sure about switching examples to ScalaFnComponent. My reasons are two-fold,

First, functional components in React JS are crap because they cannot implement `shouldComponentRender` and if you look at nearly all the scenarios in which a functional component is a good fit, you realise these would nearly always benefit from `shouldComponentRender`. React devs say that one day they might add some optimisation to make functional components faster where as currently, 1. they don't, and 2. they're crippled such that they can't be made fast. If things change in the future with React itself's handling of these then I'm likely to scratch this objection.

Second, `ScalaComponent.builder` is the default out-of-the-box means of component creation and demonstrating `ScalaFnComponent` as well is likely to make the demos (which are tutorials) less clear, a little more confusing as new users will start wondering why the inconsistency as it's unexplained.

Granted the second argument isn't super-super-strong on its own but combined with the first I'm inclined to recommend we stick with `ScalaComponent.builder` only for the demos.

What do you think?

malaman:
Thanks for the feedback.
I've updated PR:
  1) Rewrited js code in 'Hello world', 'Timer', 'Todo List', 'Refs', 'Product Table' examples without JSX, but with ES6 classes ('Refs' and 'Product table' use JSX in current master)
  2) Removed ScalaFnComponent usage in all places expect of [TodoExample.scala](https://github.com/japgolly/scalajs-react/pull/399/commits/9acb554dc7ac38b230be9503668e4512004c82bd#diff-52f62d06d60113e349035312caf4a21f). (I've decided to leave one example as a reference, because i've seen questions in gitter room regarding usage examples of ScalaFnComponent)

japgolly:
Thanks for the update.

> I've decided to leave one example as a reference, because i've seen questions in gitter room regarding usage examples of ScalaFnComponent

Good idea! :+1:

gh-pages/src/main/scala/ghpages/examples/ProductTableExample.scala
```diff
@@ -13,121 +13,134 @@ object ProductTableExample {
 
   val jsSource =
     """
-      |var ProductCategoryRow = React.createClass({
-      |    render: function() {
-      |        return (<tr><th colSpan="2">{this.props.category}</th></tr>);
```
japgolly: Oh this was in JSX before? I didn't realise that....

gh-pages/src/main/scala/ghpages/examples/ProductTableExample.scala
```diff
@@ -13,121 +13,134 @@ object ProductTableExample {
...
-      |                <tbody>{rows}</tbody>
-      |            </table>
-      |        );
-      |    }
-      |});
+      |class ProductRow extends React.Component {
+      |  render() {
+      |    var name = this.props.product.stocked ?
+      |      this.props.product.name :
+      |      <span style={{color: 'red'}}>
```
japgolly: You've got some JSX. Do you think I'm being too rigid for not enough value? ProductTableExample is a big one so maybe JSX is just better here... I'll let you decide JS or JSX for this but could you ensure that it's consistent through the example? Cheers

gh-pages/src/main/scala/ghpages/examples/RefsExample.scala
```diff
@@ -13,38 +13,43 @@ object RefsExample {
...
       |      // This code executes after the component is re-rendered
       |      this.refs.theInput.focus();   // Boom! Focused!
       |    });
-      |  },
+      |  }
       |
-      |  render: function() {
+      |  render() {
       |    return (
-      |      <div>
```
japgolly: Oh my, more existing JSX. I really thought we had none... :/

japgolly: I think I my anti-JSX arguments were wrong :( I think they hold for the first few tutorials that you see on React's site but once you're past those it looks to have been a mixture and I haven't heard anyone having problems with it. Do as you see fit. Sorry for the trouble. :)

gh-pages/src/main/scala/ghpages/examples/ProductTableExample.scala
malaman: Thanks for feedback. 
I've updated PR (ProductTableExample uses only JSX now).
I could update also all other JS examples to use JSX.

japgolly:
LGTM. :rocket: 
Thank you for the PR and your patience.

Issue #362, opened 2017-05-04
$.zoom... ?


joan38:
Hi,

I saw that we can do stuff like `$.zoom...` in this doc https://github.com/japgolly/scalajs-react/blob/master/doc/USAGE.md#react-extensions

Unfortunalty I was not able to find the `zoom` function on `$`, so either it doesn't exist anymore then the doc is outdated, or I'm missing an import or a dependency then the doc is not precis enough or I'm mistaken.

In any case I'm happy to update the doc according to the solution :)
Thanks again for this amasing library, this is awesome!

japgolly:
Oops, it looks like that snippet of doc is out of date, it's now `zoomState` instead of `zoom`.
The methods are [here](https://github.com/japgolly/scalajs-react/blob/master/core/src/main/scala/japgolly/scalajs/react/StateAccess.scala) (which all components extend, [see](https://github.com/japgolly/scalajs-react/blob/master/core/src/main/scala/japgolly/scalajs/react/component/Generic.scala#L95)).

You can also see a similar [example/demo here](https://japgolly.github.io/scalajs-react/#examples/state-snapshot) that uses Monocle.

> Thanks again for this amasing library, this is awesome!

Thank you! I'm glad to hear you're enjoying it :smiley:

joan38:
@japgolly Thanks for your answer.
But I still can't access `zoomState`, do you see what's wrong here:
```
def component[Params <: HList, ParamValues] =
    ScalaComponent
      .builder[Props[Params, ParamValues]](getClass.getSimpleName)
      .initialState {
        val taskInfo = Task.Info(runId = RunId.newId())
        (taskInfo, Map[String, Any](RunId.name -> taskInfo.runId))
      }
      .renderP { ($, props) =>
        def runTask = Callback.alert(s"""RunId: ${$.state._2}
                                        |Parameters: ${props.values($.state._2).toString}
                                        |""".stripMargin)
        //Callback(AutowireClient[Api].runTask('ahah).call())

        $.zoomState()
        val displayState = Displayer.State(kv => $.modState(s => s.copy(_2 = s._2 + kv)), key => $.state._2.get(key)) // TODO: Make it better with lenses

        <.div(
          <.div(props.name),
          props.displays(displayState),
          <.button(^.onClick --> runTask, "Run")
        )
      }
      .build
```

joan38:
[...]

PR #409, opened 2017-08-31
update docs for zoomState usage

malaman:
Addresses https://github.com/japgolly/scalajs-react/issues/362


japgolly:
Ah, good catch! Thanks

PR #416, opened 2017-09-29
Add scalajs-react-bridge to the external resources

ramnivas:


japgolly:
LGTM, thanks for sharing

Issue #421, opened 2017-10-30
Add Reusability#logNonReusable


japgolly:
Good for quick reusability debugging

```scala
  implicit class Temp[A](private val self: Reusability[A]) {
    def logNonReusable: Reusability[A] =
      logNonReusable(_.toString)

    def logNonReusable(show: A => String): Reusability[A] =
      Reusability { (a, b) =>
        val r = self.test(a, b)
        if (!r)
          println(s"Non-reusability:  ${show(a)}\n  ${show(b)}")
        r
      }
  }
```

japgolly:
Closed via #433

PR #433, opened 2018-01-01
Add Reusability#logNonReusable, addresses #421

SkyWriter:
Adds Reusability#logNonReusable wrapper to aid quick debugging.

### DIFF
--- a/doc/PERFORMANCE.md
+++ b/doc/PERFORMANCE.md
@@ -73,12 +73,10 @@ and Scalaz classes `\/` and `\&/`. For all other types, you'll need to teach it 
 * `Reusability.indexedSeq` uses `.length` and `.apply(index)` to check each element in order.
 * `Reusability.{double,float}` exist and require a tolerance to be specified.
 * `Reusability.{always,never,const(bool)}` for a hard-coded reusability decision.
-* `Reusability.logNonReusable` wraps any reusability to println a message about non-reusability to aid quick debugging.
 
 If you're using the Scalaz module, you also gain:
 * `Reusability.byEqual` uses a Scalaz `Equal` typeclass.
 * `Reusability.byRefOrEqual` uses reference equality and if different, tries using a Scalaz `Equal` typeclass.
-
 
 #### Example
 The following component will only re-render when one of the following change:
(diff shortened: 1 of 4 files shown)


SkyWriter:
@japgolly hopefully I got your idea right :-)

SkyWriter:
Wow, such a detailed feedback. Thank you for taking your time to walk me through.

I have a couple of notes/questions though:

- Just to explain my reasoning: I've deliberately placed `logNonReusable` into the object instead of the class. The class seemed to me very generic (all combinators and stuff), and method like this looked a little bit out of place. Anyways, I absolutely trust you to point out where it fits.

- How about making a title just a part of the format and not supplying it as a seprate argument? E.g. `fmt: (=> String, => String) => String = (x,y) => s"Non-reusability:\n- $x\n- $y"`.

SkyWriter:
@japgolly Apart from my question, I think I did all the changes :+1: Once again, thanks for the review!

japgolly:
No problem! In response to your questions:

1. My reasoning about class vs object is based on OO's message parsing roots, and/or how OO is implemented: (non-virtual) class methods have an implicit hidden parameter which is the class instance itself (aka `this`). Loosely speaking, I normally mentally partition method arguments into sets of mandatory and optional. If the mandatory set of args contains exactly one data type defined in the subject library/project, then I make it a class method, otherwise I make it an object method. Not just in terms of organisation but also it's also better in terms of call-site syntax. Although to reiterate, it's not a hard rule, there are cases in which other considerations exist that trump it, but it most often applies.

2. Yes, you're right that having just `fmt` instead of `title` and `fmt` is normally better practice. However there's a special use case I want to accommodate, I can imagine using this in a few places and specifying only the title, eg. `.logNonReusable(title = "Sidebar")` and `.logNonReusable(title = s"User ${user.id}")`. If users want to accomplish the same using only `fmt` you can imagine how tedious it becomes.

SkyWriter:
> My reasoning about class vs object is based on OO's message parsing roots, and/or how OO is implemented...

Right, of course! With my FP experience coming from Haskell and being in a "functional mode" I pretty often tend to forget that Scala is a fully-blown OO language too, and OO practicies apply as well.

> However there's a special use case I want to accommodate, I can imagine using this in a few places and specifying only the title ... Can you please ... introduce a title arg.

Good thinking. I've introduced the `title` parameter.

> Can you please ... reintroduce def logNonReusable that simply calls logNonReusable(). The empty parens at the end usually represents side-effects in Scala conventions so I'd like to avoid it's need in downstream users' code.

Done!

Issue #436, opened 2018-01-25
Add ^.display.grid and related attributes


joan38:
Missing: `^.display.grid`, workaround: `^.display := "grid"`
Missing: `^.gridTemplateColumns`, workaround: `Style("grid-template-columns") := "1fr 0fr 0fr 0fr"`
Missing: `^. gridTemplateRows`, workaround: `Style("grid-template-rows") := "1fr 0fr 0fr 0fr"`
Missing: `^. gridColumnStart`, workaround: `Style("grid-column-start") := "2"`
Missing: `^. gridColumnEnd`, workaround: `Style("grid-column-end") := "five"`
Missing: `^. gridRowStart`, workaround: `Style("grid-row-start") := "row1-start"`
Missing: `^. gridRowEnd`, workaround: `Style("grid-row-end") := "3"`
Missing: `^. gridColumn`, workaround: `Style("grid-column") := "3 / span 2"`
Missing: `^. gridRow`, workaround: `Style("grid-row") := "third-line / 4"`
Missing: `^. gridArea`, workaround: `Style("grid-area") := "main"`
Missing: `^. gridTemplateAreas`, workaround: `Style("grid-template-areas") := "main"`
Missing: `^. justifySelf`, workaround: `Style("justify-self") := "start"`
Missing: `^. gridTemplate`, workaround: `Style("grid-template") := ""`
Missing: `^. alignSelf`, workaround: `Style("align-self") := "start"`
Missing: `^. gridColumnGap`, workaround: `Style("grid-column-gap") := "100px 50px 100px"`
Missing: `^. gridRowGap`, workaround: `Style("grid-row-gap") := "10px"`
Missing: `^. gridGap`, workaround: `Style("grid-gap") := "10px 15px"`
Missing: `^. justifyItems`, workaround: `Style("justify-items") := "center"`
Source: https://css-tricks.com/snippets/css/complete-guide-grid/

I'm wondering something, it sounds like we are doing a hard copy from what we have in scalacss, so can't we ditch all the CSS stuff from scalajs-react and replace it with scalacss? Is it possible to define scalacss attributs directly in the tags without creating an object extending StyleSheet.Inline?

japgolly:
This would be a pretty easy PR if you have time.

[...]

PR #461, opened 2018-03-24
CSS grid styles, resolving issue #436

SkyWriter:
This is my attempt at resolving https://github.com/japgolly/scalajs-react/issues/436.

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
@@ -762,6 +762,8 @@ trait HtmlStyles {
      * according to the flexbox model.
      */
     final def `inline-flex` = this := "inline-flex"
+    /** Behaves like a grid layout container */
+    final def grid = this := "grid"
   }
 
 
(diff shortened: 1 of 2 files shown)


SkyWriter:
Hm, my previous comment seems to have disappeared somehow.

David, would you mind skimming through the draft of the code when you get a minute and tell me if I'm going in the right direction at all? I've deliberately pushed a half-baked version just so that I don't get too far in case I'm lost in the woods already.

Also, can you tell me what is project idiomatic strategy of dealing with name clashes like this one:
```
[error] /home/travis/build/japgolly/scalajs-react/core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala:2527: method alignSelf is defined twice;
[error]   the conflicting object alignSelf was defined at line 2163:10
[error]   final def alignSelf = Style[String]("align-self")
[error]             ^
```

Thank you and have a great weekend, @japgolly!

SkyWriter:
Speaking of name clashes, what I did is giving a name of `alignSelfGrid` to the grid related alignment attribute. At first I considered prefixing rather than suffixing, but it'll make it a pain to look for (unless, of course, all of the grid attributes are prefixed), so I chose the former.

core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
```diff
@@ -2393,4 +2395,178 @@ trait HtmlStyles {
      */
     final def `keep-all` = this := "keep-all"
   }
+
+  /**
+    * Grid documentation credits to Chris Coyier and a team of swell people putting
+    * together wonderful css-tricks.com website.
+    */
```
japgolly: This isn't annotating anything

core/src/main/scala/japgolly/scalajs/react/vdom/HtmlStyles.scala
```diff
@@ -2393,4 +2395,178 @@ trait HtmlStyles {
...
+    /** Fills the whole width of the grid area (this is the default) */
+    final def stretch = this := "stretch"
+  }
+
+  /**
+    * Aligns the content inside a grid item along the column axis (as opposed to
+    * justify-self which aligns along the row axis). This value applies to the
+    * content inside a single grid item.
+    */
+  object alignSelfGrid extends Style[String]("alignSelf") {
```
japgolly: No need to create a new object here. Instead the new values (below) should be added to the existing `alignSelf` object seeing as they both have the same style name.

doc/changelog/1.2.1.md
```diff
@@ -1,3 +1,28 @@
...
+    * `gridRow`
+    * `gridArea`
+    * `gridTemplateAreas`
+    * `justifySelf`
+    * `gridTemplate`
+    * `alignSelfGrid` (corresponding to the `align-self` CSS attribute)
+    * `gridColumnGap`
+    * `gridRowGap`
+    * `gridGap`
+    * `justifyItems`
```
japgolly: Good on you for updating the changelog! Can you just please sort these bullet points?

SkyWriter:
@japgolly thanks for the review! I changed everything you've mentioned.

I have a subtle feeling related to merging the `alignSelf` thing though. I would call this feeling "type unsafety" 😄. Somehow it doesn't seem right to have only slightly related values in one object.

The reason I've created a separate object is to discourage invalid combinations (e.g. having "align-self: baseline" for a grid layout). It's obvious, however, that the only _right_ way to do it is via compiler type safety. Like having a separate type for different `display` valued elements, and then in turn make them only accept certain kinds of `align-self`'ves. Have you tried pursuing this way? Or do you think it'll end being too complex to be even worth it?

japgolly:
Thanks for the changes, @SkyWriter. It looks perfect now!

Yes, you're quite right that there's no compile-time proof that the CSS or HTML attributes make sense on a given tag, nor that they make sense in conjunction with each other. Like nearly all problems of these (with the exception of literals) I imagine it's possible to do, but the machinery and the width of types would be *huuuuuge*. People already find the component types in `TYPES.md` intimidating -- the change to allow a compile-time proof of HTML attr & style correctness would make the component types look like child's play, not to mention it would probably slow compilation down by about 10x. As reverent as I am about catching and enforcing things at compile-time, it has to be balanced against the dev experience and in this case, it would be a completely unbalanced tradeoff unfortunately. Maybe in a future Scala. Nice idea though. :smile:

japgolly:
Merged! Thanks again for the contribution @SkyWriter

SkyWriter:
Oh, yeah, you're right. Types can get crazy convoluted up to a point when even compiler error messages will be a puzzle on their own. Maybe next time then.

Glad to have it merged, David! Much appreciated! And once again thank you for the hard of work of bringing the whole project to life.

PR #465, opened 2018-04-19
added onUploadProgrss to Ajax helper

an-tex:
When using the Ajax helper to upload files, the onProgress event listener (which attaches to xhr.onprogress)  doesn't trigger. This is done separately through xhr.upload.onprogress. 

I've added onUploadProgress to the Ajax helper to attach such a listener. I was thinking if it would make sense to rename the old one to onDownloadProgress but that would break the existing API so it's maybe good enough.

### DIFF
--- a/doc/changelog/1.2.1.md
+++ b/doc/changelog/1.2.1.md
@@ -2,6 +2,7 @@
 
 * Bugfix: `Reusability.shouldComponentUpdateWithOverlay` no longer crashes with React 16
 * Introduced CSS grid related attributes
+* added onUploadProgress to AjaxHelper (onProgress only triggers for the download part)
 
 
 ## Detail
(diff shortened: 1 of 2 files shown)


japgolly:
Hi @an-tex, thanks for the pull request and I'm terribly sorry about the long delay to be being able to give this attention. This PR looks great to me and bonus points for updating the changelog. I'll merge it now!

Also wrt renaming `onProgress` to `onDownloadProgress`, I think that's a fantastic idea. We needn't break the existing API but just deprecate the current method and forward it to the new one. If you're willing to do that I'd merge that too :)

an-tex:
Ha sure, I'll make another PR for the onDowloadProgress incl \@deprecated on the older one. No idea why I didn't think along that obvious path...

PR #468, opened 2018-05-24
Introduce IE hash change monitoring within the router, resolves #267

### LINKED ISSUE #267: Router not reacting on hash changes in Internet Explorer.
I have a problem with the hash based routing together with IE, when the user change manually change the url in the address bar. I'm using the router in scalajs-react 0.10.3 and scalajs 0.6.7

To easy reproduce the error, go to 
https://japgolly.github.io/scalajs-react/#examples/timer
with Internet explorer (11). The timer page is shown.

Now manually change the end of the url from /timer to /todo  and press enter. The todo page is now not shown as expected. Hit refresh and the todo page will be loaded.

This only occurs in IE (11) and not in Chrome or Firefox.

I was able to do a quick workaround, but this is probably not the most idiomatic way to solve it. 

``` scala
val (router,routerLogic) = Router.componentAndLogic(baseUrl,routerConfig)
window.addEventListener("hashchange", { (event: dom.Event) =>
  routerLogic.syncToWindowUrl.map(resolution=> routerLogic.ctl.refresh).flatten.runNow()
  event.preventDefault()
}, false)
```

Is this a know bug/feature?
Is there a better way to fix this?

SkyWriter:


japgolly:
Awesome! This looks nice and simple. The only reason I'm not immediately hitting merge is that I'm slightly concerned about the effect on non-IE. With this change do non-IE browsers now response twice to a hash change event? Do we need to change the logic to use `popstate` for non-IE and `hashchange` for IE so that it's still just one event listener in each browser?

SkyWriter:
Oops. Totally missed your comment, @japgolly, sorry about that. Let me conduct a few tests in other browsers and get back at you.

japgolly:
Hi @SkyWriter ,
I've confirmed that in Chrome this PR results in double the event being processed twice so it will need fixing.

I want to release 1.2.1 today and it'd be good to get this fix instead of waiting for 1.3.0 so I'll merge it in now and finish it off myself using this as my guide: https://github.com/elm-lang/navigation/commit/21cae6cc9e713b6ca85f0353030f95ef8b7e5267

SkyWriter:
@japgolly You're the man...



Thank you so much!

I promise the next time I'll be more helpful!

japgolly:
Hahaha thanks @SkyWriter

PR #479, opened 2018-07-20
fix broken link to how to declare an imperative

sidnt:


japgolly:
Ah it must've changed over time. Thanks!

PR #489, opened 2018-08-09
Add onCompleteHandle to Ajax

allantl:
Add `onCompleteTry` to Ajax so that its easier to pattern match against ajax response, e.g.

```
Ajax(req.method, req.url)
    .send(req.body)
    .onCompleteTry { 
      case Success(xhr) => ...
      case Failure(AjaxException(xhr)) => ....
    }
```

### DIFF
--- a/doc/changelog/1.3.2.md
+++ b/doc/changelog/1.3.2.md
@@ -1,0 +1,3 @@
+## 1.3.2
+
+* Add `onCompleteHandle` to Ajax
--- a/extra/src/main/scala/japgolly/scalajs/react/extra/Ajax.scala
+++ b/extra/src/main/scala/japgolly/scalajs/react/extra/Ajax.scala
@@ -1,7 +1,9 @@
 package japgolly.scalajs.react.extra
 
 import japgolly.scalajs.react.{Callback, CallbackKleisli, CallbackTo}
+import org.scalajs.dom.ext.AjaxException
 import org.scalajs.dom.{ProgressEvent, XMLHttpRequest}
+
 import scala.scalajs.js
 
 /** Purely-functional AJAX that runs a [[Callback]], and accepts XHR-callbacks as [[Callback]] instances.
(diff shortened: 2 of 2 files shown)


japgolly:
I'm a bit wary about having `(xhr.status >= 200 && xhr.status < 300) || xhr.status == 304` in the library, especially hardcoded, I guess it would be less concerning if it were an overridable default.

allantl:
I copied that check from scalajs's ajax (https://github.com/scala-js/scala-js-dom/blob/47af225e9f3022b9deffbbbb5dd4b098dec2beca/src/main/scala/org/scalajs/dom/ext/Extensions.scala#L319), seems that it would be sufficient for most ppl.

I can change it to take an implicit config, with that check as a default, WDYT?

japgolly:
Hi @allantl, sorry about the delay. This has been particularly hard to make a decision on to the degree that I've had long discussions with a few friends and still not arrived at a clear conclusion.

So let's just do it,  get it in and move on :) There are some minor changes to make please.

1. Instead of using `Try` (which allows *any* kind of error), please use `Either[AjaxException, XMLHttpRequest]`.
2. We'll need a new name for `onCompleteTry`. Some considerations: `onComplete` already exists and is for completion regardless of success or failure. The success/failure classification here (`success = [200,300) | 304`) is very, very general and not guaranteed to be appropriate on the server-side, for example some API calls 201 = success, all else = failure. Considering all of this what should it be called? `onGeneralSuccess`? `onSeemingSuccess`? I don't know.
3. Please update the 1.3.1 changelog instead of 1.3.0 because I released 1.3.0 just the other day.

japgolly:
Also after typing the above, I decided to model an idea I had. What would you think about doing something like this?

```scala
def blah(success: XMLHttpRequest => Boolean): (Either[AjaxException, XMLHttpRequest] => Callback) => Step2 =
  f => onComplete(xhr => f(
    if (success(xhr))
      Right(xhr)
    else
      Left(AjaxException(xhr))
  ))

def blahStatusFn(f: Int => Boolean) = blah(xhr => f(xhr.status))

def blahStatusIs(success: Int)      = blahStatusFn(_ == success)

def blahGeneric                     = blahStatusFn(s => (s >= 200 && s < 300) || s == 304)
```

so then general usage (with the `[200,300) || 304` logic) would look like this:

```scala
      Ajax(req.method, req.url)
        .setRequestContentTypeJsonUtf8
        .send(req.body)
        .blahGeneric {
          case Right(xhr) => ???
          case Left(err)  => ???
        }
```

and if one wanted to be more precise, usage would be like this:

```scala
      Ajax(req.method, req.url)
        .setRequestContentTypeJsonUtf8
        .send(req.body)
        .blahStatusIs(201) {
          case Right(xhr) => ???
          case Left(err)  => ???
        }
```

I think something like this would be a good way to easily allow the generic logic and also support more precision with nice UX and easy API. WDYT?

allantl:
Hi @japgolly,

I really appreciate you taking the time to review this.

I like the approach you mentioned in your last post.
What do you think of the name `handleResponse`?

japgolly:
Hi @allantl , again sorry for delays :)

Yeah naming is quite hard, I like the consistency of using `on` instead of `handle` but then `onResponse` seems to overlap with `onComplete` in meaning. I don't know. Maybe it's fine, probably best not to overthink :smile:

allantl:
Hi @japgolly,

I've made the changes. Let me know what you think of this.

Thanks!

japgolly:
LGTM. Thanks for getting this PR though :grinning: Cheers @allantl

japgolly:
Hi @allantl. I've refactored parts of `Ajax` now that in addition to `Callback`, 1.4.0 will have `AsyncCallback`s too. I just want to let you know that I've refactored part of this PR so the API will be slightly different now.

Instead of `.onCompleteHandle(failure | success)` the new API will be `.validateStatusIsSuccessful(failure).onComplete(success)`.

And instead of `.onCompleteHandleStatusIs(200)(failure | success)` the new API will be `.validateStatusIs(200)(failure).onComplete(success)`.

etc.

See: https://github.com/japgolly/scalajs-react/blob/series/v1.4.x/extra/src/main/scala/japgolly/scalajs/react/extra/Ajax.scala#L121-L135

japgolly:
Also here's an example of it in action: https://github.com/japgolly/scalajs-react/blob/series/v1.4.x/gh-pages/src/main/scala/ghpages/examples/AjaxExample2.scala

japgolly:
@allantl thank you and no worries! New release out soon, should be in Jan :)

And cheers, I'll fix that typo now

PR #499, opened 2018-09-20
Minor addition of missing word in README

netzwerg:
It's the little things that count, right?! :-)


japgolly:
Changes like this are always welcome! Thanks @netzwerg

PR #511, opened 2018-11-29
Change API links to point to 2.12

matthughes:
The scaladoc site is much improved in 2.12.   It would be nice if the default docs links went to that instead of the old 2.11.


japgolly:
Thanks!

PR #512, opened 2018-12-24
Fix some typos in doc

ngochai94:
Hi! I get started with scalajs-react through the docs and spotted a few typos.


japgolly:
Awesome, these kind of corrections are always welcome so thank you!

PR #517, opened 2019-01-23
Form support take2

matthughes:
Adds support for form attribute: https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#attr-fae-form


japgolly:
Great! Thanks for making those changes

PR #529, opened 2019-04-18
Make AsyncCallback.fromFuture sync if the Future it is derived from is

### LINKED ISSUE #528: `AsyncCallback.fromFuture` introduces an async boundary even if the `Future` is already completed
This causes the following component to flash "loading" on each render:

```scala
ScalaComponent
  .builder[Unit]("FromFuture")
  .render_(React.Suspense(<.span("loading"), AsyncCallback.fromFuture(Future.successful(<.span("done!")))))
  .build
```

While this one (that uses the "native" `AsyncCallback.pure`) stays synchronous and loads immediately:

```scala
ScalaComponent
  .builder[Unit]("FromFuture")
  .render_(React.Suspense(<.span("loading"), AsyncCallback.pure(<.span("done!"))))
  .build
```

I'll try to sort out a proper test case and patch ASAP.

teozkr:
Fixes #528. This is important for avoiding glitches when using Suspense with
external libraries (which often use Futures as a common interchange type for
async tasks). It's also nice to avoid the ScalaJS-React dependency in non-view
code.

### DIFF
--- a/core/src/main/scala/japgolly/scalajs/react/AsyncCallback.scala
+++ b/core/src/main/scala/japgolly/scalajs/react/AsyncCallback.scala
@@ -105,7 +105,11 @@ object AsyncCallback {
 
   def fromFuture[A](fa: => Future[A])(implicit ec: ExecutionContext): AsyncCallback[A] =
     AsyncCallback(f => Callback {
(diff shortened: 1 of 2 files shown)


japgolly:
Perfect, thanks!

PR #589, opened 2019-11-10
MonadError.raiseError for AsyncCallback was not suspending execution.

rpiaggio:
MonadError[AsyncCallback, Throwable].raiseError would raise the exception as soon as called.


japgolly:
Superseded by #590

PR #590, opened 2019-11-11
Cats-effects type class instances with law testing.

rpiaggio:
This PR includes https://github.com/japgolly/scalajs-react/pull/589 because it's necessary for this one to work, so feel free to discard the other one.

Here we provide type class instances to achieve `Sync[CallbackTo]` and `Effect[AsyncCallback]`. Also, law testing is provided.

However, 2 law tests (out of 157) are failing in `catsEffects/test`. I need help in fixing those. These are:
- `Effect[AsyncCallback].effect.repeated callback ignored *** FAILED ***`
- `Effect[AsyncCallback].effect.bracket release is called on Completed or Error *** FAILED ***`

Also, please let me know if you need other changes to make this into a publishable artifact.

Thank you!


japgolly:
Hey @rpiaggio, thanks for the work here!

* I've closed the other PR as you suggested
* Could you please also update `doc/changelog/1.5.0.md`
* The `repeated callback ignored` test is a bit of a worry. I'm not sure I like the idea of silently ignoring repeated attempts to satisfy the same async callback. I guess we could implement a check to only accept the first call in a section and issue a console warning for subsequent calls...
* As to the bracket test, I spent a surprising amount of time trying to figure out why it was failing and I believe the problem is with the `genAsync` and `genNestedAsync` generators in that they are based on arbitrary `(Try[A] => CB) => CB)` which is unlawful. A lawful `AsyncCallback` instance must call its continuation exactly once (or at least once given the above point), where as those generators are ignoring the continuation arg and just returning `Callback.empty`

cats-effects/src/main/scala/japgolly/scalajs/react/effects/AsyncCallbackEffects.scala
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  implicit val asyncCallbackDefer: Defer[AsyncCallback] = new AsyncCallbackDefer {}
+
+  trait AsyncCallbackBracket extends Bracket[AsyncCallback, Throwable] {
+    override def bracketCase[A, B](acquire: AsyncCallback[A])(use: A => AsyncCallback[B])(release: (A, ExitCase[Throwable]) => AsyncCallback[Unit]): AsyncCallback[B] =
+      acquire.flatMap { a =>
+        handleErrorWith(use(a))(t => release(a, ExitCase.Error(t)).flatMap(_ => raiseError(t)))
+          .flatMap(b => release(a, ExitCase.Completed).map(_ => b))
+      }
```
japgolly: For this bit I suggest the following. It didn't end up fixing the bracket problem but I think it's easier to read :)

```scala
      acquire.flatMap { a =>
        use(a).attempt.flatMap {
          case Right(b) => release(a, ExitCase.Completed).ret(b)
          case Left(e)  => release(a, ExitCase.Error(e)) >> AsyncCallback.throwException(e)
        }
      }
```

rpiaggio: Sounds good, done!

rpiaggio:
> Hey @rpiaggio, thanks for the work here!

My absolute pleasure.

> * Could you please also update `doc/changelog/1.5.0.md`

Please let me know if what I did is what you expected. Thank you for planning to include this in `1.5.0`. 

About the failing tests: to be honest, this is all quite new to me and I'm learning a fair bit not only of `cats-effects` but also of `discipline` and `scalacheck`.

For the `getAsync` generators, TBH, I just copied and adapted from https://github.com/typelevel/cats-effect/blob/master/laws/shared/src/main/scala/cats/effect/laws/discipline/arbitrary.scala#L39, but maybe I'm breaking laws there.

And also, TBPH, I don't quite finish grasping the laws that are failing. If you had the time to explain them to me a bit more maybe I could be of more help. If you lack the time or disposition, I can always try the `cats-effects` gitter channel I guess.

Thank you again!

japgolly:
Hey @rpiaggio. Thanks for updating the changelog, it's appreciated and looks great :+1:

Regarding that broken law we were talking about, I guess the question is whether `IO.async(_ => ())` is considered lawful, and that's a question for the cats-effect gurus cos I don't know 😛 In scalajs-react, the equivalent would be an instance of `AsyncCallback(_ => Callback.empty)` which I would say is not lawful for `AsyncCallback` because it will never complete or fail. I guess it's lawful if we want to represent an async effect that never terminates. How do those bracket laws pass if the underlying IO never terminates? I have no idea...

Also I had a look at `genAsyncF`. In IO, an `IO` instance can be sync or async but in scalajs-react they're separate types and conversions back and forth. You've got `genAsync` and you've got `genFlatMap` and those two cover the space of what `genAsyncF` would do for `IO` so I think you're good to delete that and not worry about it.

I hope that helps :)

rpiaggio:
> Hey @rpiaggio. Thanks for updating the changelog, it's appreciated and looks great 👍

Thank you.

> Regarding that broken law we were talking about, I guess the question is whether `IO.async(_ => ())` is considered lawful, and that's a question for the cats-effect gurus cos I don't know 😛 In scalajs-react, the equivalent would be an instance of `AsyncCallback(_ => Callback.empty)` which I would say is not lawful for `AsyncCallback` because it will never complete or fail. I guess it's lawful if we want to represent an async effect that never terminates. How do those bracket laws pass if the underlying IO never terminates? I have no idea...

Thank you for the explanation, I understand now. I tried to follow the similar logic in `cats-effects` and it seems to have a similar logic, but I'm not 100% sure. I must still admit not knowing how to generate a `(Try[A] => Callback) => Callback` that does anything other that returning a `Callback.empty`.

> Also I had a look at `genAsyncF`. In IO, an `IO` instance can be sync or async but in scalajs-react they're separate types and conversions back and forth. You've got `genAsync` and you've got `genFlatMap` and those two cover the space of what `genAsyncF` would do for `IO` so I think you're good to delete that and not worry about it.

Thank you for that, I removed the commented line.

SystemFw:
`async(_ => ())` is lawful (indeed it's the definition of `F.never`), and means that you get an `F` that never terminates. This makes sense when you consider that the full cats-effect machinery has a notion of _interruption/cancelation_ as well. Interruption is also the reason why you need a notion of`bracket` that isn't just `attempt.flatMap` (in general).

> The repeated callback ignored test is a bit of a worry. I'm not sure I like the idea of silently ignoring repeated attempts to satisfy the same async callback. I guess we could implement a check to only accept the first call in a section and issue a console warning for subsequent calls...

that's a viable strategy yeah

rpiaggio:
Thank you very much @SystemFw for your help.

> > The repeated callback ignored test is a bit of a worry. I'm not sure I like the idea of silently ignoring repeated attempts to satisfy the same async callback. I guess we could implement a check to only accept the first call in a section and issue a console warning for subsequent calls...
> 
> that's a viable strategy yeah

@japgolly Isn't this what `memo()` does? (minus the console warning)

Also, @SystemFw do you have any insights as to why the bracket law may be failing?

japgolly:
> Isn't this what memo() does? (minus the console warning)

@rpiaggio nah that's a bit different. With memo, the scope of "once" is universal (hence the `()` to indicate impurity). For our issue here, the scope is a single execution.

That probably doesn't make much sense does it? How about this:

```scala
val a = AsyncCallback.point{ println("hello") }
val m = a.memo()

a.toCallback.runNow() // prints hello
a.toCallback.runNow() // prints hello

m.toCallback.runNow() // prints hello
m.toCallback.runNow() // does nothing
```

Where as what we need to change to adhere to the Async laws is `AsyncCallback(k => k(Right(123)) >> k(Right(456)))` should be equivalent to `AsyncCallback(k => k(Right(123)))`

rpiaggio:
I managed to fix the double callback problem. It turns out that since tests are using `.asCallbackToFuture`, which uses an internal `Promise`, we could just change the call from `promise.complete` to `promise.tryComplete`.

The test is now working, but I didn't add logic to log when this case happens.

Now for the bracket test... Help appreciated.

rpiaggio:
I've been looking a bit more into the bracket law and I can reach the following conclusions:

* As @japgolly says, the case does seems to be that that the async is not completing.
* As @SystemFw says, an async that never completes (never calls its cb) is valid. However, should the bracket test fail in these cases? (ref to the test: https://github.com/typelevel/cats-effect/blob/93ac33dccdc2d8e7d0d7d763ba6dc3d29ee7dd20/laws/shared/src/main/scala/cats/effect/laws/AsyncLaws.scala#L74)
* Why doesn't it fail on `cats-effects` constructs?

* If I comment out the lines that generate Asyncs with callbacks, the test pass. This leads me to believe that we are not really generating good arbitraries for Asyncs. Actually, we are never generating one that does call the callback function. I cannot find how `cats-effects` achieves this and I need help putting together the pieces of `scalacheck` and `discipline` in order to achieve this.

(Note that `IO.async` receives a `(Either[Throwable, A] => Unit) => Unit`, which I can't find where it is being generated. `AsyncCallback` receives a `(Try[A] => Callback) => Callback`.)

SystemFw:
> Note that IO.async receives a (Either[Throwable, A] => Unit) => Unit, which I can't find where it is being generated. 

https://github.com/typelevel/cats-effect/blob/b4163d4c51e2207c38db84cb3de7f21ef085e76f/laws/shared/src/main/scala/cats/effect/laws/discipline/arbitrary.scala

rpiaggio:
Thank you @SystemFw !

Yes, I could find that the `IO.async` is being generated at https://github.com/typelevel/cats-effect/blob/b4163d4c51e2207c38db84cb3de7f21ef085e76f/laws/shared/src/main/scala/cats/effect/laws/discipline/arbitrary.scala#L65

But I cannot find (or follow) where a `(Either[Throwable, A] => Unit) => Unit` is generated that actually calls the callback, making the `IO.async` terminate.

rpiaggio:
I finally understood that we have to call the callback in our `Cogen[Callback]`. That makes the missing test pass.

All law tests are passing now.

rpiaggio:
Also rebased on master.

japgolly:
Oh wow, awesome @rpiaggio !! Well done!! We'll get this merged soon...

And btw, I forgot to mention it but in 1.5.0 I added `AsyncCallback.never` which is indeed lawful as @SystemFw pointed out so thanks for that. Funnily enough I actually ended up needing it a few weeks ago :laughing:

build.sbt
```diff
@@ -11,6 +11,7 @@ val monocle       = ScalajsReact.monocle
 val monocleScalaz = ScalajsReact.monocleScalaz
 val monocleCats   = ScalajsReact.monocleCats
 val cats          = ScalajsReact.cats
+val catsEffects   = ScalajsReact.catsEffects
```
japgolly: Sorry to be pedantic but I think this should be `catsEffect` as the package name is `cats.effect`

cats-effects/src/main/scala/japgolly/scalajs/react/effects/AsyncCallbackEffects.scala
```diff
@@ -0,0 +1,90 @@
...
+
+import scala.util.{Either, Failure, Success}
+
+trait AsyncCallbackEffects {
+  private val asyncCallbackMonadError: MonadError[AsyncCallback, Throwable] =
+    CatsReact.reactAsyncCallbackCatsInstance
+
+  trait AsyncCallbackDefer extends Defer[AsyncCallback] {
+    override def defer[A](fa: => AsyncCallback[A]): AsyncCallback[A] =
+      AsyncCallback.byName(fa)
```
japgolly: Not quite sure what the semantics of `Defer` are but would `AsyncCallback.lazily` be more appropriate here?

cats-effects/src/main/scala/japgolly/scalajs/react/effects/AsyncCallbackEffects.scala
```diff
@@ -0,0 +1,90 @@
...
+import cats.{Defer, MonadError}
+import japgolly.scalajs.react.{AsyncCallback, Callback, CatsReact}
+
+import scala.util.{Either, Failure, Success}
+
+trait AsyncCallbackEffects {
+  private val asyncCallbackMonadError: MonadError[AsyncCallback, Throwable] =
+    CatsReact.reactAsyncCallbackCatsInstance
+
+  trait AsyncCallbackDefer extends Defer[AsyncCallback] {
```
japgolly: Instead of having traits inside traits, could you please move inner traits into an `object AsyncCallbackEffects`?

cats-effects/src/main/scala/japgolly/scalajs/react/effects/AsyncCallbackEffects.scala
```diff
@@ -0,0 +1,90 @@
...
+    override def suspend[A](thunk: => AsyncCallback[A]): AsyncCallback[A] =
+      AsyncCallback.byName(thunk)
+  }
+
+  implicit val asyncCallbackSync: Sync[AsyncCallback] = new AsyncCallbackSync {}
+
+  trait AsyncCallbackLiftIO extends LiftIO[AsyncCallback] {
+    def liftIO[A](ioa: IO[A]): AsyncCallback[A] = {
+      AsyncCallback(cb =>
+        ioa.attempt.unsafeRunSync().fold(t => cb(Failure(t)), a => cb(Success(a)))
```
japgolly: Will this work when `IO` is doing async stuff?

cats-effects/src/main/scala/japgolly/scalajs/react/effects/CallbackToEffects.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import cats.effect.{Bracket, ExitCase, Sync}
+import cats.{Defer, MonadError}
+import japgolly.scalajs.react.{AsyncCallback, CallbackTo, CatsReact}
+
+trait CallbackToEffects {
+  private val callbackToMonadError: MonadError[CallbackTo, Throwable] =
+    CatsReact.reactCallbackCatsInstance
+
+  trait CallbackToDefer extends Defer[CallbackTo] {
```
japgolly: Same here. Can we move inner traits into the object please

doc/changelog/1.5.1.md
```diff
@@ -0,0 +1,5 @@
+## 1.5.1
```
japgolly: hehe thank you

project/Build.scala
```diff
@@ -312,6 +315,21 @@ object ScalajsReact {
...
 
+  lazy val catsEffects = project
+    .in(file("cats-effects"))
+    .configure(commonSettings, publicationSettings, extModuleName("cats-effects"))
+    .dependsOn(core, cats)
+    .settings(
+      libraryDependencies ++= Seq(
+        "org.typelevel" %%% "cats-core" % Ver.Cats,
+        "org.typelevel" %%% "cats-effect" % Ver.Cats,
+        "org.typelevel" %%% "cats-effect-laws" % Ver.Cats % "test",
```
japgolly: `s/"test"/Test/`

project/Build.scala
```diff
@@ -312,6 +315,21 @@ object ScalajsReact {
       unmanagedSourceDirectories in Compile += (sourceDirectory in (monocleScalaz, Compile)).value / "scala" / "japgolly" / "scalajs" / "react" / "internal",
       libraryDependencies += "com.github.julien-truffaut" %%% "monocle-core" % Ver.MonocleCats)
 
+  lazy val catsEffects = project
+    .in(file("cats-effects"))
+    .configure(commonSettings, publicationSettings, extModuleName("cats-effects"))
```
japgolly: here and two lines above (plus changelog): `cats-effects => cats-effect`

japgolly:
FYI next version is going to be 1.6.0 instead of 1.5.1

build.sbt
rpiaggio: Makes total sense.

cats-effects/src/main/scala/japgolly/scalajs/react/effects/AsyncCallbackEffects.scala
rpiaggio: I refactored the whole thing to have just one big `object AsyncCallbackEffects`. It will match implicit requests for all its superclasses.

cats-effects/src/main/scala/japgolly/scalajs/react/effects/AsyncCallbackEffects.scala
rpiaggio: I'm pretty sure `Defer` should not reuse. In fact, there seems to be a test for that (called "repeated sync evaluation not memoized") which fails if we use `lazily` instead of `byName`.

rpiaggio: BTW, defer tests were removed during unification. This is tested as part of `Sync` now (`Sync.suspend`).

cats-effects/src/main/scala/japgolly/scalajs/react/effects/AsyncCallbackEffects.scala
rpiaggio: Good catch. Probably will throw an exception in that case. Let me rethink this.

rpiaggio: Converted the logic to use `unsafeRunAsync`.

japgolly:
Merged! Thank you so much @rpiaggio for the contribution and all the back and forth. I'll get a 1.6.0 release out today

rpiaggio:
Thank you! My pleasure!

PR #591, opened 2019-11-16
Added queryToMap

rleibman:


test/src/test/scala/japgolly/scalajs/react/extra/router/RouterTest.scala
```diff
@@ -92,18 +105,24 @@ object RouterTest extends TestSuite {
       def assertRoot()         = testView("",          Root)
       def assertRouteHello()   = testView("/hello",    Hello)
       def assertRouteNameBob() = testView("/name/bob", Greet("bob"))
+      def assertRouteQueryParam() = testView("/queryParams?a=123&b=456&c=Hello%20bob!",
+        QueryParamPage(Map("a" -> "123", "b" -> "456", "c"-> "Hello bob!")))
```
japgolly: Great start but I think we need to test these and a combination of them together:
* A hash `#something` before or after the query params
* Empty values `eg ?x=&y=`

extra/src/main/scala/japgolly/scalajs/react/extra/router/Dsl.scala
```diff
@@ -620,6 +620,39 @@ final class RouterConfigDsl[Page] {
       if (m.matches) f(m) else None
     }
 
+  private val queryCaptureRegex = "(:?(\\?.*)|())#?|$"
```
japgolly: Pls make a `def` instead of a `val` so that Scala.JS removes it from output if it's not used.

extra/src/main/scala/japgolly/scalajs/react/extra/router/Dsl.scala
```diff
@@ -620,6 +620,39 @@ final class RouterConfigDsl[Page] {
...
+          key -> value
+        }
+        Some(params.toMap)
+      },
+      paramsMap => {
+        paramsMap.foldLeft("")((str, param) => {
+          if (str.isEmpty) {
+            s"?${param._1}=${param._2}"
+          } else {
+            str + s"&${param._1}=${param._2}"
```
japgolly: s"$str&${param._1}=${param._2}"

extra/src/main/scala/japgolly/scalajs/react/extra/router/Dsl.scala
```diff
@@ -620,6 +620,39 @@ final class RouterConfigDsl[Page] {
...
+          }
+        })
```
japgolly: This might be more readable if you do something like `_.toIterator.map(kv => s"${kv._1}=${kv._2}").mkString("?", "&", "")`

japgolly: Also, do we need to urlencode the values?

test/src/test/scala/japgolly/scalajs/react/extra/router/RouterTest.scala
japgolly: Can we also test that "Hello bob!" turns back into "Hello%20bob!"?

rleibman:
That's a lot of work... I pretty much just copied the code from the original issue and it worked for me. But you're right, they are worthy changes, I'll have time soon to get to them.

japgolly:
Yeah it's a bit more work than it seemed upfront because I want to make sure it's solid and works in all cases. The work you've done already to create the PR and add some tests is huge help and personally, pushes me over the hump for this meaning if you don't have time/will to finish it off, I'm happy to :)

japgolly:
@rleibman I added on this in #594 to test and handle all kinds of valid cases. Please have a look and let me know if you think it's good to merge

Issue #639, opened 2020-02-04
AsyncCallback is not stack-safe


fthomas:
We've been using `AsyncCallback` as effect type for a basic IndexedDB integration layer and noticed that it blows the stack for deeply nested `flatMap` calls. This happens for example when `traverse`ing a lot of IndexedDB calls for retrieving or setting values.

The issue can be reproduced with this test case:
```scala
import cats.Monad
import cats.effect.IO
import cats.implicits._
import japgolly.scalajs.react.AsyncCallback
import japgolly.scalajs.react.CatsReact.reactAsyncCallbackCatsInstance
import scala.scalajs.js.JavaScriptException
import org.scalatest.funsuite.AnyFunSuite

class StackSafetyTest extends AnyFunSuite {
  def nestedFlatMapsInTailrecLoop[F[_]](implicit F: Monad[F]): F[Int] = {
    val n = 50000
    @scala.annotation.tailrec
    def sum(list: List[F[Int]])(acc: F[Int]): F[Int] =
      list match {
        case Nil          => acc
        case head :: tail => sum(tail)(head.flatMap(h => acc.map(_ + h)))
      }
    sum(List.fill(n)(F.pure(1)))(F.pure(0))
  }

  def nestedFlatMapsInNonTailrecLoop[F[_]](implicit F: Monad[F]): F[Int] = {
    val n = 50000
    def sum(list: List[F[Int]])(acc: F[Int]): F[Int] =
      list match {
        case Nil          => acc
        case head :: tail => head.flatMap(h => sum(tail)(acc.map(_ + h)))
      }
    sum(List.fill(n)(F.pure(1)))(F.pure(0))
  }

  test("AsyncCallback is not stack-safe") {
    intercept[JavaScriptException](nestedFlatMapsInTailrecLoop[AsyncCallback].unsafeToFuture())
  }

  test("AsyncCallback does not trampoline") {
    intercept[JavaScriptException](nestedFlatMapsInNonTailrecLoop[AsyncCallback].unsafeToFuture())
  }

  test("IO is stack-safe") {
    nestedFlatMapsInTailrecLoop[IO].unsafeToFuture().onComplete(println)
  }

  test("IO trampolines") {
    nestedFlatMapsInNonTailrecLoop[IO].unsafeToFuture().onComplete(println)
  }
}
```
Withtout the `intercept[JavaScriptException]` assertions, the test fails with the following exception:
```
 - AsyncCallback is not stack-safe *** FAILED ***
[...]

PR #640, opened 2020-02-05
Stack safety for {,Async}Callback

japgolly:
Addresses #639

### DIFF
--- a/test/src/test/scala/japgolly/scalajs/react/core/CallbackTest.scala
+++ b/test/src/test/scala/japgolly/scalajs/react/core/CallbackTest.scala
@@ -7,6 +7,12 @@ object CallbackTest extends TestSuite {
 
   override def tests = Tests {
 
+    "attempt" - {
+      val t = new RuntimeException("fake error")
+      val a = CallbackTo.throwException(t).attempt.runNow()
+      a ==> Left(t)
+    }
+
     "memo" - {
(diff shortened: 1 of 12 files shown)


test/src/test/scala/japgolly/scalajs/react/core/CallbackTest.scala
```diff
@@ -50,5 +50,11 @@ object CallbackTest extends TestSuite {
     }
     */
 
+    "stackSafety" - {
+      import japgolly.scalajs.react.CatsReact._
+      type F[A] = CallbackTo[A]
+      "nestedFlatMapsInTailrecLoop"    - StackSafety.nestedFlatMapsInTailrecLoop[F]
+      "nestedFlatMapsInNonTailrecLoop" - StackSafety.nestedFlatMapsInNonTailrecLoop[F]
```
fthomas: @japgolly I just checked that if you append `.toCallback` or `.unsafeToFuture()` to both `AsyncCallback`, you'll observe the exception mentioend in #639.

japgolly: Argh! How silly of me! I should've paid proper attention to your tests instead of blindly copy-pasting and making assumptions. Thank you. That's what I was missing. It's failing as expected now.

PR #682, opened 2020-04-14
Allow pass-through properties in Router.

rpiaggio:
This allows a router to optionally receive `Props` and pass them through, as explained in the docs.

Care has been taken to maintain the previous version by means of aliasing and implicits. For example, the new router is `RouterP[Page, Props]` and there is still a `Router[Page]` which behaves as a `RouterP[Page, Unit]`.

### DIFF
--- a/doc/ROUTER.md
+++ b/doc/ROUTER.md
@@ -30,6 +30,7 @@ libraryDependencies += "com.github.japgolly.scalajs-react" %%% "extra" % "1.6.0"
   - [Setting page title](#setting-page-title)
   - [Post-render callback](#post-render-callback)
   - [Nested routes (modules)](#nested-routes-modules)
+  - [Pass-through poperties](#pass-through-properties)
 - [Examples](#examples)
 
 
@@ -56,6 +57,7 @@ Features
 * Uses HTML5 History API.
 * Callback for route changes.
 * Routing logic is easily unit-testable.
+* Optional pass-through properties (allowing, for example, passing global state to child components).
 
 
 Caution
@@ -554,17 +556,12 @@ val privatePages = (emptyRule
 
 ### Rendering with a layout
 
-Once you have a `RouterConfig`, you can call `.renderWith` on it to supply your own render function that will be invoked each time a route is rendered. It takes a function in the shape: `(RouterCtl[Page], Resolution[Page]) => VdomElement` where a `Resolution` is:
+Once you have a `RouterConfig`, you can call `.renderWith` on it to supply your own render function that will be invoked each time a route is rendered. It takes a function in the shape: `(RouterCtl[Page], Resolution[Page]) => VdomElement` where a `Resolution` provides:
 
-```scala
-/**
- * Result of the router resolving a URL and reaching a conclusion about what to render.
- *
- * @param page Data representation (or command) of what will be drawn.
- * @param render The render function provided by the rules and logic in [[RouterConfig]].
- */
-final case class Resolution[P](page: P, render: () => VdomElement)
-```
+| Member | Type | Description |
+|--------|------|-------------|
+| `page` | `Page` | Data representation (or command) of what will be drawn. |
+| `render` | `() => VdomElement` | The render function provided by the rules and logic in `RouterConfig` |
 
 Thus using the given `RouterCtl` and `Resolution` you can wrap the page in a layout, link to other pages, highlight the current page, etc.
 
@@ -680,6 +677,66 @@ If we imagine `BaseUrl` to be `http://www.example.com/` then the sitemap for `Ou
 | `http://www.example.com/#login` | `Login` |
 | `http://www.example.com/#module` | `Nest(ModuleRoot)` |
 | `http://www.example.com/#module/detail` | `Nest(ModuleDetail)` |
+
+### Pass-through properties
+
+Sometimes you don't want your `Router` to be the top-level component. In these cases, it's possible to configure the `Router` component to receive `Props`, which will be optionally forwarded to children components, the layout function, post-render functions, etc. A typical usage case for this setup is if you have a wrapping component that manages the application state.
+
+To achieve this:
+ * Write the router configuration within a `RouterConfigDslP[Page, Props].buildConfig` (instead of a `RouterConfigDsl[Page].buildConfig`).
+ * Create the router by invoking `RouterC(baseUrl, routerConfig)` (instead of `Router(baseUrl, routerConfig)`). The resulting component will accept `Props` as component properties and pass them through wherever your configuration instructs it to.
+
+##### Actions
+
+In the configuration DSL you can use the following in addition to the regular actions:
+
+| DSL | Args |
+|-----|------|
+| `renderP` | `Props => VdomElement` |
+| `renderRP` | `(RouterCtl, Props) => VdomElement` |
+| `dynRenderP` | `(Page, Props) => VdomElement` |
+| `dynRenderRP` | `(Page, RouterCtl, Props) => VdomElement` |
+
+##### Configuration
+
+Additionally, the following configurations functions are available in addition to the regular ones:
+
+| Function | Args |
+|----------|------|
+| `renderWithP` | `(RouterCtl[Page], ResolutionP[Page, Props]) => Props => VdomElement)` |
+| `setPostRenderP` | `(Option[Page], Page, Props) => Callback)` |
+| `onPostRenderP` | `(Option[Page], Page, Props) => Callback)` |
+| `setTitleP` | `(Page, Props) => String)` |
+| `setTitleOptionP` | `(Page, Props) => Option[String])` |
+
+As can be seen, `renderWithP` now provides a `ResolutionP[Page, Props]` (instead of a `Resolution[Page]`). It provides a `renderP` function which takes `Props` as parameters.
+
+Example:
+
+```scala
+case class AppState(loggedUser: String, ...)
+
+val routerConfig = RouterConfigDslP[Page, AppState].buildConfig { dsl =>
+  import dsl._
+
+  (emptyRule
+  | staticRoute(root,     Home)  ~> renderP(appState => HomePage.component(appState))
+  | staticRoute("#hello", Hello) ~> renderP(appState => <.div(s"Hello again ${appState.loggedUser}!"))
+  | staticRedirect("#hey")       ~> redirectToPage(Hello)(Redirect.Replace)
+  ) .notFound(redirectToPage(Home)(Redirect.Replace))
+    .renderWithP( (ctl, r) => appState =>
+      <.div(s"Welcome ${appState.loggedUser}")(
+        r.renderP(props)
+      )
+    )
+}
+
+val RouterComponent = RouterC(baseUrl, routerConfig)
+
+RouterComponent(appState)
+
+```
+
 
 Examples
 ========
--- a/doc/ROUTER.md
+++ b/doc/ROUTER.md
@@ -30,7 +30,7 @@ libraryDependencies += "com.github.japgolly.scalajs-react" %%% "extra" % "1.6.0"
   - [Setting page title](#setting-page-title)
   - [Post-render callback](#post-render-callback)
   - [Nested routes (modules)](#nested-routes-modules)
-  - [Pass-through poperties](#pass-through-properties)
+  - [Pass-through properties](#pass-through-properties)
 - [Examples](#examples)
 
 
@@ -683,8 +683,8 @@ If we imagine `BaseUrl` to be `http://www.example.com/` then the sitemap for `Ou
 Sometimes you don't want your `Router` to be the top-level component. In these cases, it's possible to configure the `Router` component to receive `Props`, which will be optionally forwarded to children components, the layout function, post-render functions, etc. A typical usage case for this setup is if you have a wrapping component that manages the application state.
 
 To achieve this:
- * Write the router configuration within a `RouterConfigDslP[Page, Props].buildConfig` (instead of a `RouterConfigDsl[Page].buildConfig`).
- * Create the router by invoking `RouterC(baseUrl, routerConfig)` (instead of `Router(baseUrl, routerConfig)`). The resulting component will accept `Props` as component properties and pass them through wherever your configuration instructs it to.
+ * Write the router configuration within a `RouterWithPropsConfigDsl[Page, Props].buildConfig` (instead of a `RouterConfigDsl[Page].buildConfig`).
+ * Create the router by invoking `RouterWithProps(baseUrl, routerConfig)` (instead of `Router(baseUrl, routerConfig)`). The resulting component will accept `Props` as component properties and pass them through wherever your configuration instructs it to.
 
 ##### Actions
 
@@ -703,20 +703,20 @@ Additionally, the following configurations functions are available in addition t
 
 | Function | Args |
 |----------|------|
-| `renderWithP` | `(RouterCtl[Page], ResolutionP[Page, Props]) => Props => VdomElement)` |
+| `renderWithP` | `(RouterCtl[Page], ResolutionWithProps[Page, Props]) => Props => VdomElement)` |
 | `setPostRenderP` | `(Option[Page], Page, Props) => Callback)` |
 | `onPostRenderP` | `(Option[Page], Page, Props) => Callback)` |
 | `setTitleP` | `(Page, Props) => String)` |
 | `setTitleOptionP` | `(Page, Props) => Option[String])` |
 
-As can be seen, `renderWithP` now provides a `ResolutionP[Page, Props]` (instead of a `Resolution[Page]`). It provides a `renderP` function which takes `Props` as parameters.
+As can be seen, `renderWithP` now provides a `ResolutionWithProps[Page, Props]` (instead of a `Resolution[Page]`), which in turn provides a `renderP` function which takes `Props` as parameters.
 
 Example:
 
 ```scala
 case class AppState(loggedUser: String, ...)
 
-val routerConfig = RouterConfigDslP[Page, AppState].buildConfig { dsl =>
+val routerConfig = RouterWithPropsConfigDsl[Page, AppState].buildConfig { dsl =>
   import dsl._
 
   (emptyRule
@@ -726,15 +726,14 @@ val routerConfig = RouterConfigDslP[Page, AppState].buildConfig { dsl =>
   ) .notFound(redirectToPage(Home)(Redirect.Replace))
     .renderWithP( (ctl, r) => appState =>
       <.div(s"Welcome ${appState.loggedUser}")(
-        r.renderP(props)
+        r.renderP(appState)
       )
     )
 }
 
-val RouterComponent = RouterC(baseUrl, routerConfig)
+val RouterComponent = RouterWithProps(baseUrl, routerConfig)
 
 RouterComponent(appState)
-
 ```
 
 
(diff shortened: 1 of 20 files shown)


japgolly:
Hey @rpiaggio , thanks a lot for this PR. That you've even updated the doc is very much appreciated. I'm pretty busy for the next few days so I'll give this a proper look sometime soon. From a very quick skim I saw one place where it's not backwards-compatible and that's the `RouterU` and `RouterM` aliases. Maybe there's more hiding somewhere too (?).

What are your thoughts on having two separate packages? One for router the way it is now and the other where it has `Props` in the API, (and of course the non-prop package would just be forwards and type aliases to the with-props package). I don't know that it's a good idea, but it's something to consider. What do you think?

Also out of curiosity, what's your use case for this? I'll merge no matter what you say so don't feel you have to justify yourself, I'm honestly just curious. My thoughts for not implementing it myself is that:

1. You'd wouldn't have multiple routers in the same app
2. It's always going to be the top-level component (with the exception of an error handler) - it adds only functionality and nothing visual
3. To pass in global settings that you configure before you initialise React, I put a (non-router) top-level component in a class (instead of an object) and pass in all info the constructor, then the router wraps the component and that component renders the global layout and delegates the inside depending on the current route.

rpiaggio:
Hi @japgolly! Thank you for your quick feedback.

I'm going to update the PR keeping `RouterU` and `RouterM` as they were before and defining new `RouterPU` and `RouterPM`. Thanks for spotting this! There might be other cases that have slipped through the cracks. Also, I think it might make sense to be able to provide `RouterP` with a `Reusability[Props]`, so I will include this.

About having two separate packages:  I don't have a formed opinion. I was trying not to introduce breaking changes or code duplication and I'd like not to need extra imports in client code. If this can be achieved, then sure, it might help to organize parts of the code. 

About the use case: we've been experimenting on different ways how to manage global application state in a FP way without using a global `var` or React's `Context`. Although the current approach looks promising at the moment, it might change drastically and we are open to suggestions. But basically:
* There's a top-level component that keeps the application state as the component state. What we are using for the moment for this is https://github.com/rpiaggio/crystal/blob/master/js/src/main/scala/crystal/react/AppRoot.scala.
* We want to pass this state (or parts of it) to the `Router`'s layout (eg: show logged user in the navbar) and the components it decides to render. Also, the state can inform decisions made in `setTitle`, `onPostRender`, etc.

BTW, this is kind of orthogonal, but in order to allow downstream components to modify the global state, we are using something similar to a `StateSnapshot`, but it differs in that the state modification is not applied to the snapshot value but it's delegated to the top-level component's `modState`. This allows different components down the tree to make updates without stepping on each other. You can see bare-minimum (I would even call it prototype) code for this here: https://github.com/rpiaggio/crystal/blob/master/shared/src/main/scala/crystal/View.scala.

rpiaggio:
I'm backing away from having `Reusability[P]` as a parameter in `Router(P)` methods. If needed, reusability can be added by calling `componentUnbuilt`. I've rearranged the methods so it can be called in both scenarios.

japgolly:
Hey @rpiaggio. Reading your comment it sounds like you and I are doing very, very similar things. So I've built a few Big Serious Things in scalajs-react, and the biggest and most complex project is my startup [ShipReq](https://shipreq.com). ShipReq is 100% FP, frontend and backend, and in it, I've got a single component that does only 2 things: 1) Delegate to a component according to current route. 2) It's stateful and contains *all state for the entire SPA*. The rest of the app is all stateless components that all accept zoomed-in `StateSnapshot`s. Literally that and some lenses to remove boilerplate from zooming-in, is all I use for state management of the biggest and most complex SPA I've ever seen or worked on for money. It's so easy, simple, even beautiful, that I really don't understand all these frameworks that try to offer you management of it like Redux etc. They're all inferior versions of a really, really simple problem from my perspective. Anyway, I'm introducing this background not to say "hey man shuttup i know what I'm doing, I know best", but to say "hey man, I've got similar goals and I did it this way and it works really, really well for me including at scale" :) Now with that background in place, let me start actually responding...

The way I did it was to have the Router above my stateful top component. I don't need to pass it any props. Instead my `RouterConfig` is a function that takes in a `Global` which is a mutable class with hidden vars in it and access to the underlying data via `Px` and/or `CallbackTo` which is safe. I guess that's the suboptimal part you want to avoid right? Don't want to define vars anywhere? Even I just realised for the first time, "oh that means I have state in *two* places! Not just that one stateful component". So in your case you're going to have the stateful component *above* the router and pass in varying data to it, right? So `AppRoot` would sit above your router. That makes sense and please let me know how it works it. It sounds like it has a lot of promise.

About not using `StateSnapshot` that one makes me sad haha. Maybe that's why my subconscious demanded I give my background above because it's just worked out so well and so beautifully for me. I basically consider calling `setState` more than once in a event handler an anti-pattern. In my big huge project I handle so much cross-page functionality and state manipulation and haven't ever come across a case where I have two completely isolated, unrelated parts of the code responding to the same event and making disjoint, unrelated changes. Again I'm not trying to say you're wrong, just that in my experience it's an anti-pattern that you're much better off addressing directly and merging into one `modState`. And what I say "better off" I also mean for the general codebase health in that things that are actually related are made explicitly related which is very good for readability and preventing future bugs. Are you sure this behaviour is absolutely necessary? Have you tried integrating the multiple `setState` calls? Also the reason that `StateSnapshot` set/mod state applied to the current value is basically one of cache invalidation and correctness. If you're ever worried that a component is on-screen having been rendered by a `StateSnapshot` with a stale value that's a huge bug! One shouldn't accept and workaround that, one should fix whatever behaviour is resulting in that happening. Unless you're doing anything really, really hacky, (in my experience) a bug in a `Reusability` instance would be the only cause for the bug. Or maybe passing `Reusable` functions around with buggy `Reusability`? I know this is all a bit tangential to the PR but I really hope my input is helpful and maybe you can get back onboard with `StateSnapshot` *and* have a cleaner, safer codebase. :smile: If your case is truly unique and my suggestion isn't valid, that's fine too of course but I'd be very curious and would like to learn why. Maybe it could lead to an improvement in scalajs-react itself :D

doc/ROUTER.md
```diff
@@ -30,6 +30,7 @@ libraryDependencies += "com.github.japgolly.scalajs-react" %%% "extra" % "1.6.0"
   - [Setting page title](#setting-page-title)
   - [Post-render callback](#post-render-callback)
   - [Nested routes (modules)](#nested-routes-modules)
+  - [Pass-through poperties](#pass-through-properties)
```
japgolly, line 33: typo: poperties => properties

doc/ROUTER.md
```diff
@@ -681,6 +678,66 @@ If we imagine `BaseUrl` to be `http://www.example.com/` then the sitemap for `Ou
 | `http://www.example.com/#module` | `Nest(ModuleRoot)` |
 | `http://www.example.com/#module/detail` | `Nest(ModuleDetail)` |
 
+### Pass-through properties
```
japgolly, line 681: Btw thank you so much for writing all this doc. It's a necessity and I find it very tedious. I appreciate it coming in with the PR <3

doc/ROUTER.md
```diff
@@ -681,6 +678,66 @@ If we imagine `BaseUrl` to be `http://www.example.com/` then the sitemap for `Ou
...
+
+Sometimes you don't want your `Router` to be the top-level component. In these cases, it's possible to configure the `Router` component to receive `Props`, which will be optionally forwarded to children components, the layout function, post-render functions, etc. A typical usage case for this setup is if you have a wrapping component that manages the application state.
+
+To achieve this:
+ * Write the router configuration within a `RouterConfigDslP[Page, Props].buildConfig` (instead of a `RouterConfigDsl[Page].buildConfig`).
+ * Create the router by invoking `RouterC(baseUrl, routerConfig)` (instead of `Router(baseUrl, routerConfig)`). The resulting component will accept `Props` as component properties and pass them through wherever your configuration instructs it to.
```
japgolly, line 687: How about:
* `RouterConfigDslP` => `RouterWithPropsConfigDsl`
* `RouterC` => `RouterWithProps`

I basically regret my "Haskell-envy" days. I don't think single letter suffixes are a good idea anymore.

japgolly: I know there's still a bunch of single-letter suffixes in scalajs-react already, but I hope to change that in future and I don't want to be adding more now if possible.

doc/ROUTER.md
```diff
@@ -681,6 +678,66 @@ If we imagine `BaseUrl` to be `http://www.example.com/` then the sitemap for `Ou
...
+##### Actions
+
+In the configuration DSL you can use the following in addition to the regular actions:
+
+| DSL | Args |
+|-----|------|
+| `renderP` | `Props => VdomElement` |
+| `renderRP` | `(RouterCtl, Props) => VdomElement` |
+| `dynRenderP` | `(Page, Props) => VdomElement` |
+| `dynRenderRP` | `(Page, RouterCtl, Props) => VdomElement` |
```
japgolly, line 698: These suffixes are fine because we already do a similar thing with `*render*` everywhere. It's consistent in that sense.

doc/ROUTER.md
```diff
@@ -681,6 +678,66 @@ If we imagine `BaseUrl` to be `http://www.example.com/` then the sitemap for `Ou
...
+
+Additionally, the following configurations functions are available in addition to the regular ones:
+
+| Function | Args |
+|----------|------|
+| `renderWithP` | `(RouterCtl[Page], ResolutionP[Page, Props]) => Props => VdomElement)` |
+| `setPostRenderP` | `(Option[Page], Page, Props) => Callback)` |
+| `onPostRenderP` | `(Option[Page], Page, Props) => Callback)` |
+| `setTitleP` | `(Page, Props) => String)` |
+| `setTitleOptionP` | `(Page, Props) => Option[String])` |
```
japgolly, line 710: On the fence about whether a `P` suffix here is better than `WithProps` suffix. WDYT?

doc/ROUTER.md
```diff
@@ -681,6 +678,66 @@ If we imagine `BaseUrl` to be `http://www.example.com/` then the sitemap for `Ou
...
+        r.renderP(props)
+      )
+    )
+}
+
+val RouterComponent = RouterC(baseUrl, routerConfig)
+
+RouterComponent(appState)
+
+```
```
japgolly, line 738: Fantastic!

Just please remove the trailing whitespace line at the end of the snippet

extra/src/main/scala/japgolly/scalajs/react/extra/router/Router.scala
```diff
@@ -9,36 +9,58 @@ import japgolly.scalajs.react.vdom.VdomElement
...
+    componentUnbuiltC[Page, Props](baseUrl, cfg, new RouterLogic(baseUrl, cfg))
+
+  def componentUnbuiltC[Page, Props](baseUrl: BaseUrl, cfg: RouterConfigP[Page, Props], lgc: RouterLogic[Page, Props]) =
+    ScalaComponent.builder[Props]("Router")
       .initialStateCallback   (lgc.syncToWindowUrl)
       .backend                (_ => new OnUnmount.Backend)
-      .render_S               (lgc.render)
-      .componentDidMount      ($ => cfg.postRenderFn(None, $.state.page))
-      .componentDidUpdate     (i => cfg.postRenderFn(Some(i.prevState.page), i.currentState.page))
+      .render                 ($ => lgc.render($.state)($.props))
```
japgolly, line 44: Can we uncurry `lgc.render` pls

extra/src/main/scala/japgolly/scalajs/react/extra/router/Router.scala
```diff
@@ -164,7 +186,7 @@ final class RouterLogic[Page](val baseUrl: BaseUrl, cfg: RouterConfig[Page]) ext
     }
   }
 
-  def render(r: Resolution): VdomElement =
+  def render(r: Resolution): Props => VdomElement =
```
japgolly, line 189: Pls put `Props` inside param list (i.e. uncurry)

extra/src/main/scala/japgolly/scalajs/react/extra/router/RouterConfigP.scala
```diff
@@ -83,8 +98,8 @@ case class RouterConfig[Page](rules       : RoutingRules[Page],
       val msg = s"${errors.size} RouterConfig errors detected:$es"
       dom.console.error(msg)
 
-      val el: VdomElement =
-        <.pre(^.color := "#900", ^.margin := "auto", ^.display := "block", msg)
+      val el: Props => VdomElement =
+        _ => <.pre(^.color := "#900", ^.margin := "auto", ^.display := "block", msg)
```
japgolly, line 102: Nein. The point here is to avoid re-creation each time.

extra/src/main/scala/japgolly/scalajs/react/extra/router/package.scala
```diff
@@ -1,12 +1,26 @@
...
+  type Router [P]         = ScalaComponent              [Unit, ResolutionP[P, Unit], OnUnmount.Backend, CtorType.Nullary]
+  type RouterU[P]         = ScalaComponent.Unmounted    [Unit, ResolutionP[P, Unit], OnUnmount.Backend]
+  type RouterM[P]         = ScalaComponent.MountedImpure[Unit, ResolutionP[P, Unit], OnUnmount.Backend]
+
+  type RouterConfig[P] = RouterConfigP[P, Unit]
+  type Resolution[P]   = ResolutionP[P, Unit]
+
+  implicit class ResolutionOps[P](resolution: Resolution[P]) {
+    def render(): VdomElement = resolution.renderP(())
+  }
```
japgolly, line 22: Better than a value class here would be to add this directly to `ResolutionP` and then ask for evidence that `P =:= Unit`. It's how the Scala collections handle `.flatten` too.

rpiaggio:
Hi @japgolly! Thank you a lot for your in depth review!

Yes, it seems that we are trying to solve the exact same problem of global state management.
What you described in your 2nd paragraph in exactly what I'm trying to achieve here: global state in a single place, which can be the `State` of a React component, and propagation using exclusively the React machinery. The router can tap into this state to make decisions and pass partial views to its children and the layout.

I will rebase, address your comments and push another commit.

--------------

Sorry to make you sad about not using `StateSnapshot`, maybe there's something I'm not getting about how to use it. First of all, I said before, the `RouterWithProps` (yes, I like that name better) could be used with `StateSnapshot`, it's an independent concern (that's why I drew a line above this paragraph).

I did try to use `StateSnapshot` but the issue I ran into is when zooming into it. For example, say we have a `case class Record(a: A, b: B)` and a component with a form to edit this `Record` that receives a `StateShanpshot[Record]`. When it renders, it call components to edit `a` and `b`, passing `StateSnapshot[A]` and `StateShanpshot[B]` by `zoomState`ing. For this to work, both subcomponents would have to rerender whenever the `Record` changes, even if the value they are showing hasn't changed. This is because `StateSnapshot` wraps  a `SetFn`, which isn't reusable in the zoomed-in `StateSnapshot`s when the base value changes. Since it's using a `SetFn`, it applies the `zoomState` setter to the wrapped value. The `View` class is heavily inspired by `StateSnapshot` but wraps a `ModFn` instead of a `SetFn`. The `ModFn` is always reusable and it's applied directly to the state where the `View` originated, not to the wrapped one, even across `zoom`s. Maybe there's a way to achieve this with `StateSnapshot` which I just haven't figured out. I'm actually very curious about how you deal with propagating and narrowing down a `StateSnapshot` when passing it down the component hierarchy.

doc/ROUTER.md
rpiaggio: Good catch. Fixed.

doc/ROUTER.md
rpiaggio: You're welcome. I was just following the https://github.com/japgolly/scalajs-react/blob/master/CONTRIBUTING.md guidelines ;)

doc/ROUTER.md
rpiaggio: Yes, I will follow your suggestion. I was trying to follow the conventions of the library, but I like that you're steering it into fuller names.

doc/ROUTER.md
rpiaggio: I think that adding `WithProps` here would be a tad too verbose. Also `renderWithWithProps` might be confusing.

extra/src/main/scala/japgolly/scalajs/react/extra/router/Router.scala
rpiaggio: 👍

extra/src/main/scala/japgolly/scalajs/react/extra/router/Router.scala
rpiaggio: 👍

doc/ROUTER.md
rpiaggio: 👍

doc/ROUTER.md
rpiaggio: 👍

extra/src/main/scala/japgolly/scalajs/react/extra/router/RouterConfigP.scala
rpiaggio: I see, changed it back and added the ignored `Props` parameter to the caller.

extra/src/main/scala/japgolly/scalajs/react/extra/router/package.scala
rpiaggio: Much nicer, thank you for the suggestion.

japgolly:
@rpiaggio I've responded to that `StateSnapshot` discussion in #707

doc/ROUTER.md
japgolly: Wow! Even that deserves a thank you hahahaha

doc/ROUTER.md
japgolly: Yeah ok no problem. I'll accept it now and just add it to the already-half-full bucket of renaming I should probably do in the future :laughing:

extra/src/main/scala/japgolly/scalajs/react/extra/router/RouterWithPropsConfig.scala
```diff
@@ -7,55 +7,70 @@ import japgolly.scalajs.react.{Callback, CallbackTo}
...
   /**
    * Change the document title after the router renders.
    *
    * @param f Given the current page that just rendered, return a new title.
    */
-  def setTitle(f: Page => String): RouterConfig[Page] =
+  def setTitleP(f: (Page, Props) => String): RouterWithPropsConfig[Page, Props] =
+    setTitleOptionP((p, c) => Some(f(p, c)))
+
+  def setTitle(f: Page => String): RouterWithPropsConfig[Page, Props] =
```
japgolly, line 60: Ah whoops I didn't catch this the first time around. Could you please copy/paste the ScalaDoc to all of these new methods, and/or make sure the old methods retain it.

rpiaggio: Ooops, sorry about the ScalaDocs. Should be fixed now. Also, did another rebase on master. Hope Travis is happy now.

japgolly:
Merged. Thanks a lot for your work and patience @rpiaggio !

rpiaggio:
> Merged. Thanks a lot for your work and patience @rpiaggio !

Thank ***you*** for your all your work in this project and your patience with this issue.

PR #703, opened 2020-05-06
Upgrade to scalajs-env-jsdom-nodejs 1.1.0.

sjrd:
This allows to re-enable the unit test for `lifecycle1` that had been moved to a manual test in 9eb8abefb0daed72fc208d5b9aaa652aaedad534.


japgolly:
Fantastic! Thanks as always @sjrd!!

A small note: I can't merge this as is because I want to keep that `js-sandbox` stuff around. I nearly always need to reverse engineer everything that React does to properly support it because the official doc isn't specific enough. I'm so sick of it being a pain each time that I'm just going to keep `js-sandbox` around so that I've got something easy to work with / hack on.

So if you like you can modify the PR to retain the sandbox files, or I'm happy to close this PR and take care of it myself. No need for you to do anything as you've done so much already :D

sjrd:
Thanks. I'll close this PR and let you do it then. It will be quicker than if we need another round.

I'm so glad this issue is finally really with.

PR #920, opened 2021-05-31
React 17

japgolly:


core/src/main/scala/japgolly/scalajs/react/ReactDOM.scala
```diff
@@ -47,4 +47,7 @@ object ReactDOM {
   @deprecated("Import vdom and use ReactPortal()", "")
   def createPortal(child: NotAllowed, container: Any) = child.result
 
+  def flushSync[A](a: CallbackTo[A]): CallbackTo[A] =
```
japgolly, line 50: Not actually new in React 17, just a new undocumented function I discovered that expects occasionally use with hooks.

PR #928, opened 2021-06-11
Effect generalisation

japgolly:


rpiaggio:
Hi David, thank you so much for addressing this. It is certainly a huge global refactor.

I've been looking at the changes. Though I know it's probably ongoing, I have some questions. Please bear in mind that I'm not familiar or don't have fresh in my mind every use case provided by the framework, so I may be missing obvious things here. In any case, the idea is to start a discussion. All of these concerns are geared towards easily using (cats-effect) `IO` as an effect.

* Do we need `Sync`/`Async` to be a `Monad`? Sure, it's probably desirable and users will want to compose them, but does the framework need this? In other words, does the framework ever compose them?

* I would love to be able to just return `IO` in event handlers (and in `useEffect` and lifecycle callbacks). Thing is, we cannot run `IO`s synchronously, so we can't define an `UnsafeSync` instance for them. What we are doing now is triggering the `IO`s from within a `Callback`, which starts it concurrently by "firing and forgetting". It isn't exactly the same thing if the framework ever needs to wait for completion, but this seems to be working quite well.

* To run an `IO` (the fire and forget I mentioned above) we need a `Dispatcher`, which we can provide from context at the call site, but we don't have it when building a general effect typeclass for `IO`. In other words, we cannot provide a run method in the effect typeclass. But if the current effect typeclasses were split in two, by having another typeclass responsible for running effects, we could instantiate that one implicitly at the call site.

* It would be nice to have `set/modState` effects that complete once the action is done and `flatMap` them in the usual way, instead of forcibly having to pass a callback as a 2nd argument. We do this by wrapping in (cats-effect) `Async` (https://github.com/rpiaggio/crystal/blob/master/js/src/main/scala/crystal/react/implicits/package.scala#L51) but we would lose this capability, unless the framework provides a way to build this idiom.

japgolly:
Hey @rpiaggio how's it going? Cool, feedback is very welcome and I'm not a cats-effect expert.

Firstly, let me give you a quick overview of how I see this working:
* There will be a version of the core module that literally doesn't even have any Callback class stuff on the classpath
* Methods that accept a callback argument will accept any type of format so long as there's an `Effect.Sync` typeclass available
* Code that returns a Callback / AsyncCallback will instead return a `DefaultEffects.Sync` / `DefaultEffects.Async`
* The `DefaultEffects` types will be customisable. `core` will extend `core-general` with the `callback` module and override the `DefaultEffects` types to be `Callback` / `AsyncCallback`. (This means that it will remain compatible with today's model for users who are happy with it.) I plan to provide another module that overrides `DefaultEffects` to be `IO`/`IO` for cats-effect users so that no manual customisation is needed, simply choosing the appropriate Maven artifacts will be enough.
* All classes that return default effects will also have `.withEffect[F]` and `.withAsyncEffect[F]` that allow users to change the effect types in an ad-hoc fashion.

Now with that context in place, let me answer the questions.

1) For `Sync`, yes, the library needs monadic ops for some functions. There was even a case where I needed `map` on `Async` yesterday but I ended up just wimping out (because I'd have to write an impl for `Async.Untyped`) and converting to `js.Promise` instead ([see here](https://github.com/japgolly/scalajs-react/blob/1b6bbc941479d67105d9109ce36c3f88102b80c3/core-general/src/main/scala/japgolly/scalajs/react/React.scala#L90-L91)). Also FYI, providing instances of these typeclasses will be something I do in scalajs-react; cats-effect users won't have to implement themselves.

2) Firstly we'd want a `Sync`, not an `UnsafeSync` for synchronous `IO` (small detail). What do you mean by this though: *"we cannot run IOs synchronously"*? Isn't it just `.unsafePerformIO()` or similar? Or is the problem that `IO` could be async and in those cases it would be a runtime effect to attempt running it synchronously? If so we should be able to get away with replacing `runSync` with a fire-and-forget method in `UnsafeSync` (which is the super-class of `Sync`). Actually, off the top of my head the when you provide `shouldComponentUpdate` callbacks a fire-and-forget wouldn't work because you need that `Boolean` result synchronously. Still, it might be an idea to change as much as possible to use a fire-and-forget approach, and just limited the space in which an async IO could be run as sync and fail (hopefully only `shouldComponentUpdate` and not more). I mean if we literally don't have any other possible alternative, there's nothing we can do anyway.

3) Oh man. That makes more sense for JVM but does it make much sense for JS? Could we just provide a static typeclass impl in the cats-effect module that uses the default cats-effect `Dispatcher` (seeing as I don't understand what the difference between different `Dispatcher`s could be in JS land). Also are you talking cats-effect v2, v3, or both?

4) That also ready exists by using appending `Async` to those methods. Eg. `setState(s, f)` ≈ `setStateAsync(s) >> f.toAsyncCallback`. When both the default `Sync` and `Async` classes as `IO`, `{set,mod}State{,Option}Async` will do exactly what you're proposing :)

rpiaggio:
Hey @japgolly. Thanks for the explanation of how things will look, sounds good! I guess you meant `SyncIO`/`IO` in your 4th bullet point.

About the numbered points:

1. Ok 👍 BTW, what are the `Untyped` types for?
2. Yes, I was referring to the `def runSync[A] (fa: => F[A]) : A` method in `UnsafeSync`(which `Sync` inherits). There's no way to implement this for `IO` in JS, since you cannot synchronously wait for async effect completion. However, after I wrote my original message I realized that using async effects generally for lifecycle/events is not a good approach because:
  * As you say, some lifecycle methods have to pass a result back to React (at least `shouldComponentUpdate` does, as you pointed out).
  * I just found out that React actually batches when multiple calls to `setState` are performed in a sync handler. This is something we definitely want to take advantage of, and for this to work React has to know when the handling is done; thus, a sync effect is needed.
  So, I'll probably refactor all of our code to switch to sync unless async is really needed (sigh). So, it might be helpful to have some convenient way to trigger ("fire and forget")an async effect from a sync context, in case we want to program generically.
3. Sorry I didn't specify the cats-effect version. I was talking about v3, which we are using throughout, including GPP. The way we do it now is the `Dispatcher` is obtained effectfully at the start of the application and we store it in context, which basically most of components then use (they can obtain it from the context, but in a lot of places we just pass it implicitly in props). I guess we could store it statically, but anyway we would need it to create the typeclass instances for `Sync` and `Async` (which seems would conflict with having global instances in `DefaultEffects`). To overcome this, I suggest using separate typeclasses to model the actual effects and the runners. The `run*` methods from the effect typeclasses would need an implicit runner, but it's not required until that point. Furthermore, this would allow you to use different dispatchers for different things if anyone wanted to do that.
4. Nice! Thank you.

japgolly:
Hey @rpiaggio 

1. The `Untyped` types are there as a secret implementation detail so that I can make certain constructions generically. The typeclasses should never be used (and I'm going to remove them soon), just the types should be used and only internally.
2. Yeah you're right. I'll look into using fire-and-forget where it makes sense.
3. I'd really prefer not to require an abstract dispatcher typeclass. To me that's a big leaky abstraction. Concerns:
    * Users would have to have a dispatcher-like implicit in scope every time they create a component. I think it's much better to define it once globally. If users want to use different dispatchers sometimes, they could also bring their own `Sync` implicit into scope (which overrides the default).
    * Would there ever be a reason to have different dispatchers in JS? On the JVM I assume a dispatcher is valuable because you can have different thread pools and work queues but on the JS everything is single-threaded. I could be wrong of course but I don't see a reason for more than a global instance. There are only two possibilities: run now or run via setTimeout, and that distinction doesn't make sense in terms of React because React has it's own effect management.
    * WDYT?
4. My pleasure! Re `SyncIO` vs `IO`, I didn't realise `SyncIO` existed. Is it new in v3? I can't find any documentation on it, I can only see the source code and it's not a super-type of `IO`. As a cats-effect user would you use `SyncIO` directly for all of your synchronous effects?

japgolly:
Oh re (3), did you just mean moving `runSync` to a new typeclass? That would be less leaky that what I first thought you meant which was to have a new typeclass for an abstract dispatcher, and then require both Sync and AbstractDispatch in order to run effects.

rpiaggio:
> Oh re (3), did you just mean moving `runSync` to a new typeclass?

Yes, that's what I meant.

`SyncIO` was there in CE2. It's basically equivalent to `CallbackTo`. This is the CE2 docs, I can't find the CE3 ones: https://typelevel.org/cats-effect/docs/2.x/datatypes/syncio. The more I think about using `IO` as a sync effect, the more I think it's not that good of an idea. React does take advantage of having sync callbacks.

japgolly:
Hey @rpiaggio,  moving `runSync` to a new typeclass is certainly possible but atm it would be a huge refactor. First I need to finish the current huge refactor I'm in the middle of and when all that is working let's give it a try as is. I still have hope that we can make it work as is so I'll ask to you have another look when done but if it doesn't work for you, np we'll refine it further and introduce that new typeclass if that's better. I'll ping you again later when I've got something more presentable to show you.

rpiaggio:
Hi @japgolly sorry for the late reply.

After some more thought, I don't think it's necessary to create a new typeclass. Sync effects should have a way of running them directly. Treating an async effect as sync doesn't seem like it would work.

japgolly:
Hey @rpiaggio, all good. I did end up creating a new typeclass `Dispatch` ([see here](https://github.com/japgolly/scalajs-react/blob/3f4f458ac5fe39407fdddf05452a1c20e01f0212/util/src/main/scala/japgolly/scalajs/react/util/Effect.scala#L17-L19)) cos there are actually plenty of places where it doesn't matter whether the callback is sync or async, just need to fire and forget.

japgolly:
Hey @rpiaggio. This isn't finished yet but it's good enough for a look. Have a look at these two scenarios:

1) With [these modules](https://github.com/japgolly/scalajs-react/blob/topic/effectGeneralisation/downstream-tests/build.sbt#L140-L142), it's the same as it is today except that you can also use `SyncIO` in place of any `CallbackTo`, and `SyncIO` or `IO` in place of any `Callback`. [Example](https://github.com/japgolly/scalajs-react/blob/topic/effectGeneralisation/downstream-tests/js/src/main/scala/downstream/Carrot.scala#L24-L26)

2) If you use the new `core-cats-effect` module instead of `core` [like this](https://github.com/japgolly/scalajs-react/blob/topic/effectGeneralisation/downstream-tests/build.sbt#L167-L168), then the `Callback` classes won't even be on the classpath, and `SyncIO` will be the default sync effect and `IO` will be the default async effect  (i.e. `$.setState` will return a `SyncIO` and `$.setStateAsync` will return an `IO`). [Example](https://github.com/japgolly/scalajs-react/blob/topic/effectGeneralisation/downstream-tests/js-ce/src/main/scala/downstream/Catnip.scala#L13-L14)

WDYT? Happy with this result?

rpiaggio:
I won't be able to look at it in depth just yet (disconnected vacation) but it sounds awesome. Thank you!

japgolly:
@rpiaggio no worries, thanks & enjoy your holiday mate!

PR #949, opened 2021-07-27
cats-effect 3.2.0. CB/IO core bundle.

rpiaggio:
Attempt to build a core bundle that uses `Callback` for sync effect and `IO` for async.

Also adjusts for some changes introduced in cats-effect 3.2.0:
* If `IOApp` was used, it would throw an exception attempting to install the global `IORuntime`, since `EffectCatsEffect.io` would have installed it already, so I made the latter a `lazy val` instead of an `object`.
* Tests for `Async[AsyncCallback]` now require a `Cogen[Output[AsyncCallback, Throwable, A]]`, which can be derived from a `Cogen[AsyncCallback[A]]`, which is now provided.

BTW: I don't like copying `ReactCallbackExtensions`.  I tried to move that file to another project that `coreBundleCallback` and `coreBundleCBIO` can depend on, but that doesn't quite work since `ReactCallbackExtensions` needs to have `Callback` defined as `DefaultEffect.Sync`.


japgolly:
> BTW: I don't like copying `ReactCallbackExtensions`.  I tried to move that file to another project that `coreBundleCallback` and `coreBundleCBIO` can depend on, but that doesn't quite work since `ReactCallbackExtensions` needs to have `Callback` defined as `DefaultEffect.Sync`.

Are you sure about that? I don't see `DefaultEffect` referenced in the `ReactCallbackExtensions` file.

japgolly:
Ok cool, apart from the copy of `ReactCallbackExtensions` everything looks good :)

There are a few others changes that we need but it shouldn't take look to do. Could you please also
* Update `doc/MODULES.md` (don't worry about the diagrams)
* Add a new module to `downstream-tests` to test the defaults and also that the non-defaults are still accepted. To test it do a `+publishLocal` from scalajs-react and then `downstream-tests/run` (or you can just run `bin/ci` which will do it automatically). Actually if you want I can do this step if this task is a bit big/confusing

rpiaggio:
> > BTW: I don't like copying `ReactCallbackExtensions`.  I tried to move that file to another project that `coreBundleCallback` and `coreBundleCBIO` can depend on, but that doesn't quite work since `ReactCallbackExtensions` needs to have `Callback` defined as `DefaultEffect.Sync`.
> 
> Are you sure about that? I don't see `DefaultEffect` referenced in the `ReactCallbackExtensions` file.

I might not have been very clear.

The extension method 
``` scala
    def asEventDefault(e: ReactEvent): CallbackOption[A] =
      (self <* e.preventDefaultCB.toCBO).unless(e.defaultPrevented)
```
needs `e.preventDefaultCB` to return a `Callback`. `preventDefault` is an extension method itself which returns a `DefaultEffect.Sync[Unit]`. Maybe you can see another way around this ?(without type-parametrizing`ReactEvent` with the effect)

rpiaggio:
> Actually if you want I can do this step if this task is a bit big/confusing

I'll give it a shot and let you know

japgolly:
> Maybe you can see another way around this ?(without type-parametrizingReactEvent with the effect)

What about this? :smiling_imp: 

    def asEventDefault(e: ReactEvent): CallbackOption[A] =
      (self <* CallbackOption(e.preventDefault)).unless(e.defaultPrevented)

project/Dependencies.scala
```diff
@@ -10,7 +10,7 @@ object Dependencies {
 
     // Externally observable
     val cats                  = "2.6.1"
-    val catsEffect            = "3.1.1"
+    val catsEffect            = "3.2.0"
```
japgolly, line 13: FYI I upgraded to 3.2.1 on `master`, thanks for the `Cogen` impl. Now this PR can be just about the bundle.

rpiaggio:
@japgolly Followup on the downstream tests:

I defined a module there but I think there's something wrong with `withRenderedIntoDocumentAsync`. When I use it (and return an `IO` from within), I get a runtime exception:
```
  org.scalajs.linker.runtime.UndefinedBehaviorError: java.lang.ClassCastException: IO(...) is not an instance of scala.Function1
    <jscode>.$throwClassCastException(main.js:53)
    <jscode>.$as_F1(main.js:15011)
    japgolly.scalajs.react.test.ReactTestUtils$$anon$7.apply(main.js:23671)
    <jscode>.{anonymous}()(main.js:40322)
    scala.scalajs.runtime.AnonFunction0.apply(main.js:38530)
```

In the current commit I'm running with `Future` anyway, and it passes as it is.

The test is not complete, but could you confirm this is along the lines that you were expecting?

tests/src/test/scala/japgolly/scalajs/react/test/TestTest.scala
```diff
@@ -253,7 +253,7 @@ object TestTest extends TestSuite {
       assert(body1 == body2)
     }
 
-    "withRenderedIntoDocumentAsync" - {
+    "withRenderedIntoDocumentFuture" - {
```
rpiaggio, line 256: I looked for a use of `withRenderedIntoDocumentAsync` but couldn't find one. Decided to rename this test.

downstream-tests/run
```diff
@@ -18,7 +18,7 @@ all_tests_values=(
   "-Ddownstream_tests.fullOptJS -Ddownstream_tests.reusability.dev=overlay"
 )
 all_tests_keys=()
-declare -A all_tests
+declare -a all_tests
```
rpiaggio, line 21: Does this still work for you? `-A` doesn't seem to work in Mac.

PR #1004, opened 2021-11-03
Derive cats.Monad+Monoid for Effect.Sync.

rpiaggio:
- Derive `cats.MonadThrow[F]` when there's an implicit `Effect.Sync[F]`.
- Derive `cats.Monoid[F[A]]` when there's an implicit `Effect.Sync[F]` and `Monoid[A]`.
- Added `throwException` to `Dispatch` typeclass.


japgolly:
:boom: 
:rocket:

PR #1013, opened 2021-12-05
WIP - Double reusability implicit

JamesMcIntosh:
WIP - I've still to test this change locally.

I wasn't sure if there was a reason for excluding an implicit Double, I've needed it a couple of times recently so thought I'd have a quick dig in the code to see where the other implicits came from.

### DIFF
--- a/coreGeneric/src/main/scala/japgolly/scalajs/react/Reusability.scala
+++ b/coreGeneric/src/main/scala/japgolly/scalajs/react/Reusability.scala
@@ -238,6 +238,7 @@ object Reusability extends ReusabilityMacros with ScalaVersionSpecificReusabilit
   @inline implicit def short  : Reusability[Short  ] = by_==
   @inline implicit def int    : Reusability[Int    ] = by_==
   @inline implicit def long   : Reusability[Long   ] = by_==
+  @inline implicit def double : Reusability[Double ] = by_==
   @inline implicit def string : Reusability[String ] = by_==
   @inline implicit def date   : Reusability[Date   ] = by_==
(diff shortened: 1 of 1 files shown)


japgolly:
Hey @JamesMcIntosh , thanks for the PR.

Unfortunately there is a reason that `Reusability[Double]` isn't provided out-of-the-box, it's not uncommon to have a bit of tolerance when comparing doubles. In fact there's a [`Reusability.double`](https://github.com/japgolly/scalajs-react/blob/master/coreGeneric/src/main/scala/japgolly/scalajs/react/Reusability.scala#L138-L139) which takes a tolerance. Eg. if some AJAX call returns 4.005134978165 and then on next call returns  4.005134971648 the majority of components would not need a re-render but having an out-of-the-box `Reusability[Double]` instance of silently just cause a re-render. Not having an implicit instance makes it a userland decision, and not just that, ensures that when userland decisions are made, they're applied as intended (as opposed to accidentally forgetting an import and having the default silently kick in).

japgolly:
For `java.time` stuff, users can import `Reusability.TemporalImplicitsWithoutTolerance._`. Maybe we should add a `Reusability.NumbersWithoutTolerance._` import too?

japgolly:
Closing in favour of #1029 which adds `Reusability.DecimalImplicitsWithoutTolerance`

JamesMcIntosh:
@japgolly Thanks for explaining the reasoning behind `Reusability.double(tolerance)` and creating the new `DecimalImplicitsWithoutTolerance`

japgolly:
@JamesMcIntosh np and thanks for raising this. It's released in 2.0.1 now so enjoy!

PR #1059, opened 2022-04-18
Add react key to SVG attributes

triggerNZ:
key was missing from SVG attributes, requiring users to use the HTML one, which is a bit awkward.


japgolly:
Hey man, sorry for delay. Caught covid, was completely fucked all of last week. Only finally out of bed for a day or so.

CI is running now but I'll bet ya a round it's gonna complain...

japgolly:
Oh dammit I guess I owe you a 🍺, `HtmlTags` and `SvgTags` aren't composed anywhere it seems, was expecting a failure due to conflicting `final val key` instances. Alright, in it goes...

japgolly:
v2.1.1 released with this fix

japgolly:
Hey man. Ack but I'm sick with covid atm, gimme another day or two.

PR #1060, opened 2022-04-25
fix: Correct the clipPath tag

nghuuphuoc:
The current version renders the `clipPathTag` tag which isn't correct.

Screen Shot 2022-04-25 at 15 06 04


japgolly:
Whoops, thank you!

japgolly:
v2.1.1 released with this fix

PR #1069, opened 2022-06-01
[WIP] Support React Refresh

### LINKED ISSUE #1061: Support `react-refresh` for faster dev experience

japgolly:
Closes #1061


library/project/Dependencies.scala
```diff
@@ -47,6 +47,7 @@ object Dependencies {
     val microlibsCompileTime = Def.setting("com.github.japgolly.microlibs" %%% "compile-time"                            % Ver.microlibs)
     val microlibsTestUtil    = Def.setting("com.github.japgolly.microlibs" %%% "test-util"                               % Ver.microlibs)
     val microlibsTypes       = Def.setting("com.github.japgolly.microlibs" %%% "types"                                   % Ver.microlibs)
+    val microlibsUtils       = Def.setting("com.github.japgolly.microlibs" %%% "utils"                       % Ver.microlibs)
```
japgolly, line 50: 😅

japgolly:
Have just run into https://github.com/lampepfl/dotty/issues/15435 on the Scala 3 side. This is effectively a blocker because workarounds would need to be applied at call-site (i.e. users' code). Worse is that there's no way to detect when we're running up against this bug; it will just silently generate incorrect code. 🤢

japgolly:
Note to self: add as JS RR test with custom hooks to confirm how it's handled in plain JS

japgolly:
Reminder: Try again to reproduce the Scala 3 lazy val bug now that I *fiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiiinally* know it's related to `betaReduction`

japgolly:
Create a Scala.js issue to elide those useless `props.a;` statements

Issue #1081, opened 2023-01-17
A minimal setup fails with Scala 3


Jake6329:
I'm new to scalajs-react, and I've set up a minimal project. It works well with Scala 2.13.10 but it fails with Scala 3.2.1 (3.1.3 also shows the same error). When I run the command `sbt "Compile / fastOptJS / webpack"` in order to bundle my app, I get this error messages:

```
...
[info] Writing scalajs.webpack.config.js
[info] Bundling the application with its NPM dependencies
[error] /Users/jakelim/demos/electron/my-scalajs-app/target/scala-3.2.1/scalajs-bundler/main/node_modules/source-map/lib/source-map-generator.js:298
[error]       throw new Error('Invalid mapping: ' + JSON.stringify({
[error]       ^
[error] Error: Invalid mapping: {"generated":{"line":13410,"column":89},"source":"webpack://https:/raw.githubusercontent.com/japgolly/scalajs-react/v2.1.1/coreGeneric/src/main/scala/japgolly/scalajs/react/ReactCaughtError.scala","original":{"line":12,"column":-1},"name":null}
[error]     at SourceMapGenerator.SourceMapGenerator_validateMapping [as _validateMapping] (/Users/jakelim/demos/electron/my-scalajs-app/target/scala-3.2.1/scalajs-bundler/main/node_modules/source-map/lib/source-map-generator.js:298:13)
[error]     at SourceMapGenerator.SourceMapGenerator_addMapping [as addMapping] (/Users/jakelim/demos/electron/my-scalajs-app/target/scala-3.2.1/scalajs-bundler/main/node_modules/source-map/lib/source-map-generator.js:110:12)
[error]     at /Users/jakelim/demos/electron/my-scalajs-app/target/scala-3.2.1/scalajs-bundler/main/node_modules/source-map/lib/source-node.js:351:13
[error]     at SourceNode.SourceNode_walk [as walk] (/Users/jakelim/demos/electron/my-scalajs-app/target/scala-3.2.1/scalajs-bundler/main/node_modules/source-map/lib/source-node.js:230:9)
[error]     at SourceNode.SourceNode_walk [as walk] (/Users/jakelim/demos/electron/my-scalajs-app/target/scala-3.2.1/scalajs-bundler/main/node_modules/source-map/lib/source-node.js:226:13)
[...]

PR #1083, opened 2023-05-18
Point at library subfolder

dispalt:
When everything moved to the subdirectory I think it broke the mapping, I was playing around with something and just noticed it stopped correctly mapping sources.

### DIFF
--- a/doc/changelog/next.md
+++ b/doc/changelog/next.md
@@ -4,3 +4,4 @@
 * Scala.js upgraded to 1.10.0
 * microlibs upgraded to 4.1.0
 * Scala 3 upgraded to 3.1.2
+* Fixed the source map uri to point to the correct location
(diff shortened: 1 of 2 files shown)


nkgm:
+1
Also discussed in #1081

dispalt:
Thoughts @japgolly ?

cquiroz:
It seems this should be fine. Did you publish locally to test?

dispalt:
No, I will do that and report back here

dispalt:
Okay tested it end to end, and glad I did because I had the version and prefix flipped.  It looks good now though.

cquiroz:
Cool, thanks a lot. Could you a add a little not on the changelog?

dispalt:
Done

cquiroz:
Thanks a lot @dispalt

PR #1089, opened 2024-01-27
Reusability in functional components via React.memo instead of class component

rpiaggio:
`shouldComponentUpdate` hook and `renderWithReuse` use an embedded class component to achieve reusability.

This PR changes the approach to the recommended use of `React.memo` for this purpose. This avoids having to mix functional and class components.

Furthermore, I'm having trouble making class components work with React 18. This unlocks the use of React 18 for apps with purely functional components.


cquiroz:
Would it make sense to make a partial release for React 18 with only functional components support?

cquiroz (approved):
LGTM

rpiaggio:
> Would it make sense to make a partial release for React 18 with only functional components support?

I think so. That's what I'm going for.

PR #1092, opened 2024-03-03
Support displayName in Scala functional components

rpiaggio:
- `ScalaFnComponent` and `ScalaForwardRef` now support `.withDisplayName(name)`. Must be called first (before hooks/`render*`/`withChildren*`).

Example:

```scala
  ScalaFnComponent.withDisplayName("MyComponent")
    .withHooks[Props]
    .useState(0)
    .render((_, s) => s.value.toString)
```

~~I disabled Mima checks temporarily since this is binary incompatible (it is source compatible though). This implies we have to step up to 2.2.0. We have to reenable Mima tests after 2.2.0 release.~~ This will go into a `3.0.0` release, together with React 18 support.

Thanks to @hugo-vrijswijk for the idea of using `sourcecode.FullName` for the default names.


library/tests/src/test/scala/japgolly/scalajs/react/core/ScalaFnComponentTest.scala
```diff
@@ -13,7 +13,7 @@ object ScalaFnComponentTest extends TestSuite {
 
   final case class Add(x: Int, y: Int)
 
-  val CaseClassProps = ScalaFnComponent[Add] { a =>
+  val CaseClassProps = ScalaFnComponent.withDisplayName("Add")[Add] { a =>
```
hugo-vrijswijk, line 16: What happens when not setting `displayName`? Does it also set the JS facade to `""`? And if so, is this different behaviour than leaving it unset?

rpiaggio: If not set, it is taken from an implicit `sourcecode.FullName`.

doc/changelog/2.2.0.md
```diff
@@ -0,0 +1,14 @@
+## 2.2.0
+
+### New Stuff
+
+- `ScalaFnComponent` and `ScalaForwardRef` now support `.withDisplayName(name)`. Must be called first (before hooks/`render*`/`withChildren*`).
```
toddburnside, line 5: Maybe something about the default value?

toddburnside (approved):
A lot of work to weave the name through everything, but I don't see a simpler alternative. LGTM.

doc/changelog/2.2.0.md
rpiaggio: Good idea 👍

rpiaggio:
Since this implementation breaks binary compatibility, it would actually imply a major version bump (ie: `3.0.0`).

Seems overkill just for this change, so I'm going to merge this change to our React 18 branch and release this feature there.

PR #1094, opened 2024-07-10
Actually invoke init in components stuck in "first"

rpiaggio:
Components that only used the `useEffect` family of hooks didn't invoke the hooks.


cquiroz (approved):
LGTM

PR #1098, opened 2024-12-11
Allow custom route parsers

rpiaggio:
This PR allows users to define custom `Route`s which turn `Path`s into pages and viceversa, without requiring them to be build via the DSL. Of course, the old DSL still works.

Also: add flake.nix, update dependencies.


cquiroz (approved):
LGTM

PR #1101, opened 2024-12-16
Monadic hooks

rpiaggio:
Implements an alternative syntax for hooks, in the form of:

``` scala
val component = ScalaFnComponent[Int] { props =>
  for {
    state       <- useState(props)
    _           <- useEffect(state.modState(_ + 1))
    forceUpdate <- useForceUpdate
  } yield
    <.div(
      state.value,
      <.button(^.onClick --> forceUpdate, "Click me")
    )
}
```

This provides a simpler syntax, while still guaranteeing safeguards to honor the rules of hooks: hook invocations are delayed and can only be invoked by `ScalaFnComponent`.

Source compatibility is preserved with the builder-style hooks.


library/ghpages/src/main/scala/ghpages/examples/HooksExample.scala
```diff
@@ -40,6 +39,16 @@ object HooksExample {
   import japgolly.scalajs.react.vdom.html_<^._
   import org.scalajs.dom.document
 
+  // Monadic hook example
+  // def useAddReporter(initial: Int): HookResult[Reusable[Int => Callback]] ={
```
toddburnside, line 43: Did you want to keep this commented code?

toddburnside (approved):
The syntax is a great improvement. It will be easier to use and much easier to create custom hooks. 🎊 Should `Hooks.md` be updated? We probably need the existing documentation for awhile for existing code, but it could be at the bottom and marked as legacy? I don't see why someone would want to develop new code with the old syntax once this is available. 😄

library/ghpages/src/main/scala/ghpages/examples/HooksExample.scala
rpiaggio: Nope, just removed it, thank you

rpiaggio:
> The syntax is a great improvement. It will be easier to use and much easier to create custom hooks. 🎊 Should `Hooks.md` be updated? We probably need the existing documentation for awhile for existing code, but it could be at the bottom and marked as legacy? I don't see why someone would want to develop new code with the old syntax once this is available. 😄

Since this a beta branch, I was planning to give this a bit more of a test before officially documenting it, in case there are changes. After we have it in the wild for some time, we can decide how to document both styles.

library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/react18.scala
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
```
japgolly, line 2: Whoops, this should be removed.

library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/HookResult.scala
```diff
@@ -0,0 +1,25 @@
...
+object HookResult {
+  @inline def apply[A](a: => A): HookResult[A] =
+    new HookResult(() => a)
+
+  implicit class HookOps[I, O](private val h: I => HookResult[O]) extends AnyVal {
+    def contramap[A](f: A => I): A => HookResult[O] = f andThen h
+  }
+
+  implicit def autoLift[A](a: => A)(implicit f: A => VdomNode): HookResult[VdomNode] =
+    HookResult(f(a))
```
japgolly, line 22: Ah so this is how users can still create `ScalaFnComponent` instances without hooks. Clever! Makes me think we should rename `HookResult` then because hooks are not always involved. Maybe something like `FnCompBody` (or similar but less ugly)?

japgolly (changes requested):
I feel like a monadic approach was considered in the early days but there was a significant disadvantage to it. Something about safety that was lost (meaning there was a React-imposed requirement that I was able to enforce at the compiler-level with the existing approach, but would become a potential user-error with a mondaic approach)? I don't remember with enough detail. Looking at the code though, this monadic approach looks pretty good. In fact, it looks very nice. 👍

Impressive!

Only other two concerns are 1) documentation update (as you've already mentioned), and 2) binary backwards-compatibility — could I hear your position on it breaking? How concerned are you? Why / why not?

library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/react18.scala
rpiaggio: Yup, sorry. I tested this out first on one of our projects and then copied the file over.

library/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "jsdom": "^25.0.1"
+  }
+}
```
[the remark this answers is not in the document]
rpiaggio: You have to have `jsdom` installed in npm for tests to run. I thought it practical to have this here and just `npm install` in order to be able to run tests.

library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/HookResult.scala
rpiaggio: The driving thought for naming this was to have a nice signature for custom hooks, which are now defined so:

```scala
def myHook(input: Foo): HookResult[Bar]
```

In the case of rendering without hooks, the name of `HookResult` doesn't leak to the user, so I thought it was ok. But you are right in that it's not precise. Maybe we can use `Output`, which seems to fit both cases and is still kind of nice to read in code?

Custom hooks would then look like:
```scala
def myHook(input: Foo): Output[Bar]
```

Or, we can even have a type alias `type HookResult[A] = Output[A]`.

rpiaggio:
> I feel like a monadic approach was considered in the early days but there was a significant disadvantage to it. Something about safety that was lost (meaning there was a React-imposed requirement that I was able to enforce at the compiler-level with the existing approach, but would become a potential user-error with a mondaic approach)? I don't remember with enough detail.

I've been trying to remember if we considered this, but can't remember either. We probably considered it but didn't consider delaying evaluation, which is what makes it safe.

That said, you can still break the rules (eg: `fold`ing over an `Option` would allow you call hooks conditionally). But I think it's still safer than free-form and it's a small price to pay for the reduced syntax.

> Looking at the code though, this monadic approach looks pretty good. In fact, it looks very nice. 👍

Thank you!

> Only other two concerns are 1) documentation update (as you've already mentioned), 

I'll add to the PR. Not sure how to handle the availability of both styles.

> and 2) binary backwards-compatibility — could I hear your position on it breaking? How concerned are you? Why / why not?

If we make this part of a major release (`3.0.0`), it wouldn't be unexpected to have a breaking change. Since it's still source-compatible, this would have a very minor impact since it would just require recompiling downstream libraries with no need for code changes.

rpiaggio:
I ended up renaming to `Delayed`, which actually describes what the type class is doing, and added a `HookResult` type alias to implement a nice API.

I also documented the new style as the default style for hooks, making it explicit that there are ways around the rules; while keeping the builder-style as an alternative that provide complete safety while being more verbose.

doc/HOOKS_BUILDER.md
```diff
@@ -0,0 +1,649 @@
...
+
+scalajs-react aims to provide a DSL which enforces usage of React's rules for
+hooks at compile time. There's actually 2 DSL flavors implemented:
+  * A composition-based approach documented [here](HOOKS.md), which results in 
+    more concise code, but doesn't protect you from breaking a couple of rules
+    via `fold`/`traverse`.
+  * A builder-like approach documented in this page, which gives you complete
+    safety at the cost of being more vebose.
+
+If you have a spare 9 hours (!), you can watch the livestreamed coding sessions
```
japgolly, line 27: 🤣🤣🤣

doc/HOOKS.md
```diff
@@ -337,322 +204,50 @@ import japgolly.scalajs.react._
...
-  val hook2: CustomHook[String, Int] = ???
-  val hooks: CustomHook[(Long, String), (Boolean, Int)] = hook1 ++ hook2
-}
-
-object Example4 {
-  val hook1: CustomHook[Unit, Boolean] = ???
-  val hook2: CustomHook[String, Unit] = ???
-  val hooks: CustomHook[String, Boolean] = hook1 ++ hook2
+  val Component = ScalaFnComponent[Unit]( _ =>
+    ExampleHook.useTitleCounter // <--- usage
```
japgolly, line 208: I'm thinking better to demonstrate this with a for-comp for consistency and clarity.

doc/HOOKS.md
```diff
@@ -337,322 +204,50 @@ import japgolly.scalajs.react._
...
-        <.p(s"You clicked $count times"),
-        <.button(
-          ^.onClick --> Callback(setCount(count + 1)),
-          "Click me"
-        ),
-      )
-    }
-}
+val newCustomHook: CustomHook[I, O] = CustomHook.fromHookResult(useHook1)
+val newCustomHook: CustomHook[Unit, O] = CustomHook.fromHookResult(useHook2)
```
japgolly, line 252: Very nice facilitating and demonstrating interop here 👍

library/coreGeneric/src/main/scala/japgolly/scalajs/react/component/Delayed.scala
```diff
@@ -0,0 +1,93 @@
+package japgolly.scalajs.react.component
+
+import japgolly.scalajs.react.vdom.VdomNode
+
+class Delayed[+A](protected[react] val eval: () => A) extends AnyVal {
```
japgolly, line 5: Ah this clarifies something for me: this is exactly the same in spirit as `CallbackTo` — the functionality is the same. It is also not stack-safe. Could we drop this in favour of `CallbackTo`?

library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/package.scala
```diff
@@ -0,0 +1,9 @@
+package japgolly.scalajs.react
+
+import japgolly.scalajs.react.component.Delayed
+
+package object hooks {
+  // Offers nicer API for hooks
+  type HookResult[A] = Delayed[A]
```
japgolly, line 7: missing covariance

japgolly (changes requested):
reviewed, feedback above

library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/package.scala
rpiaggio: good catch

library/coreGeneric/src/main/scala/japgolly/scalajs/react/component/Delayed.scala
rpiaggio: Yes! I thought of using `CallbackTo` initially, but I'm not sure we want to interop. Also, I agree it should be stack stafe.

I'll explore if we can unify code and report back.

rpiaggio: Well, we can't directly use `CallbackTo` without a major reshuffle, since `CallbackTo` is not available in `core`.

I tried implementing via `DefaultSync`, but found out we can't use `DefaultSync` in `core` because it becomes hard-linked.

So, I moved `Trampoline` to `util` and implemented `Delayed` in terms of `Trampoline`, just like `CallbackTo` does. Anyway, we probably don't want a lot of the thiings that `CallbackTo` provides, like timing, prompts, etc. This way we can ensure the minimal functionality we actually need here, plus not mixing the types, without involving `Delayed` with generic effects.

japgolly: Ok that sounds like a perfect trade-off. Thanks for working through that.

japgolly (approved):
LGTM. Nice work!

PR #1104, opened 2024-12-26
Implement missing React 18 hooks, useCallback accepts non-unit afunctions

rpiaggio:
Implements:
- `useDeferredValue`
- `useSyncExternalStore`
- `useInsertionEffect`

`useCallback` now accepts functions that return a sync value (before it only worked on functions that returned sync or async `Unit`).

### DIFF
--- a/library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/Api.scala
+++ b/library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/Api.scala
@@ -6,6 +6,7 @@ import japgolly.scalajs.react.feature.Context
 import japgolly.scalajs.react.hooks.Hooks._
 import japgolly.scalajs.react.internal.Box
 import japgolly.scalajs.react.util.DefaultEffects
+import japgolly.scalajs.react.util.Effect.Sync
 import japgolly.scalajs.react.util.Util.identityFn
 import japgolly.scalajs.react.vdom.{TopNode, VdomNode}
 import japgolly.scalajs.react.{Children, CtorType, Ref, Reusability, Reusable}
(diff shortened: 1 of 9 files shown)


library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/Api.scala
```diff
@@ -335,13 +336,103 @@ object Api {
...
+    final def useInsertionEffectBy[A](effect: Ctx => A)(implicit a: UseEffectArg[A], step: Step): step.Self =
+      self(ctx => UseEffect.unsafeCreateInsertion(effect(ctx)))
+
+    /** The signature is identical to [[useEffect]], but it fires synchronously after all DOM mutations, but before any 
+      * layout Effects fire. Use this to insert styles before any Effects fire that may need to read layout. Updates 
+      * scheduled inside useLayoutEffect will be flushed synchronously, before the browser has a chance to paint.
+      *
+      * Prefer the standard [[useEffect]] when possible to avoid blocking visual updates.
+      *
+      * If you'd only like to execute the effect when your component is mounted, then use [[useInsertionEffectOnMount]].
```
toddburnside, line 382: I think the documentation for these hooks was copy/pasted. 😄 For example, this IS `userInsertionEffectOnMount`.

toddburnside (approved):
Wow. That was a lot typing to add those. 😆 LGTM

library/coreGeneric/src/main/scala/japgolly/scalajs/react/hooks/Api.scala
rpiaggio: Good catch, I fixed the `use*Effect*` Scaladocs  throughout now.

PR #1106, opened 2024-12-29
Update act

rpiaggio:
Starting in React 18.3, `act` has been moved to `React` and deprecated in `ReactTestUtils`.

Also, this PR implements 3 methods in `ReactTestUtils2`:
- `renderAsync`, which invokes `act(root.render(component))` asnychronouslly;
- `actAsync(=> A)` which is a convenience method for async `act` without the need of wrapping the action in an async effect;
- `withRenderedAsync`, which is an async version of `withRendered`.

This enables async testing with the pattern:
```scala
withRenderedAsync(comp()){ _ =>  // Mounts comp in root
  assert(...)
  for
    _ <- actAsync(Simulate.click(...))
    _ =  assert(...)
    _ <- actAsync(Simulate.click(...))
  yield assert(...)
}.map(_ => assert(...))          // After unmount
```

`act` will become fully async in the future. The current React docs read:
```
We recommend using act with await and an async function. Although the sync version works in many cases, it doesn’t work in all cases and due to the way React schedules updates internally, it’s difficult to predict when you can use the sync version.

We will deprecate and remove the sync version in the future.
```

Note: This PR subsumes https://github.com/japgolly/scalajs-react/pull/1105, since React 18.3 is needed.

### DIFF
--- a/library/testUtil/src/main/scala/japgolly/scalajs/react/test/ReactTestUtils2.scala
+++ b/library/testUtil/src/main/scala/japgolly/scalajs/react/test/ReactTestUtils2.scala
@@ -11,9 +11,6 @@ import org.scalajs.dom.html.Element
 import org.scalajs.dom.{console, document}
 
 object ReactTestUtils2 extends ReactTestUtils2 {
-
-  val raw = japgolly.scalajs.react.test.facade.ReactTestUtils
-
   IsReactActEnvironment = true
 
   private[ReactTestUtils2] object Internals {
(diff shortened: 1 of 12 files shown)


library/testUtil/src/main/scala/japgolly/scalajs/react/test/ReactTestUtils2.scala
```diff
@@ -113,12 +114,27 @@ trait ReactTestUtils2 extends japgolly.scalajs.react.test.internal.ReactTestUtil
     WithDsl(newElement())(removeElement)
 
   val withReactRoot: WithDsl[TestReactRoot, ImplicitUnit] =
-    withElement.mapResourse(TestReactRoot(_))(_.unmount())
+    withElement.mapResource(TestReactRoot(_))(_.unmount())
```
cquiroz, line 117: nice 😄

cquiroz (approved):
LGTM

PR #1109, opened 2025-04-20
remove some implicits

cquiroz:
I noted the family of functions `mkTag...` take the function implicitly.
I'm not sure it is needed. In this PR the implicit modifier is removed


cquiroz:
This needs some rework

rpiaggio:
There are some useful implicits around for converting eg: `String` into `VdomNode`. The implicit parameter lets you take advantage of that.

PR #1115, opened 2025-08-29
Update sbt-ci-release and sbt

rpiaggio:
We may need to remove the suffix from the organization, but this may work as it is so I suggest we try it first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ metals.sbt
 node_modules/
 package-lock.json
 target
+.direnv/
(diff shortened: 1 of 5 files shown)


japgolly:
Thanks! Merged. Trying a release: https://github.com/japgolly/scalajs-react/actions/runs/17334806884

japgolly:
Closer, but unfortunately that didn't work `gigahorse.StatusError: Unexpected status: 401`

rpiaggio:
IIRC, you have to migrate the credentials to the central server in sonatype.

rpiaggio:
https://dev.to/mxro/migrating-maven-namespace-to-central-portal-fgp

